PR #49, opened 2026-06-12
Fix OHLCV cache

thomasdfowler:
Fix OHLCV cache treating overnight/weekend gaps as stale, causing yfinance refetch on every request.


JohnFunkCode (approved):
Good catch! Reviewed change, validated on my machine!

PR #71, opened 2026-07-02
Fix OHLCV corruption: serialize yf.download (not thread-safe)

thomasdfowler:
## Incident
Prod OHLCV was corrupted twice (2026-06-30 ~20:10 UTC, 2026-07-02): 200+ symbols' daily bars rewritten with **other tickers' data**, in byte-identical groups of exactly 4 (e.g. `INTC=MRVL=POWL=SOLS`, `GOOG=MSFT=MU=PATH` — full OHLCV+volume identical). Symptom in the UI: wrong/blank prices in the Securities view; screener returned nonsense (`MSFT last_close=10.68`).

## Root cause
`refresh_options_snapshots` (wired to the UI **Refresh All** button) fetches chains from a `ThreadPoolExecutor(max_workers=4)`. Each fetch computes Bollinger bands → `get_history` → on a stale cache, `_fetch_yfinance` runs `yf.download` **concurrently in 4 threads**. `yf.download` is not thread-safe — results pass through module-global state (`yfinance.shared._DFS`), so concurrent calls read each other's frames. `_store_bars` then upserts the wrong ticker's history. `max_workers=4` ⇒ corrupted rows in groups of 4. Caught live on 2026-07-02: `CORRECTED bar detected: AMZN cached=240.14 new=26.57` with NET receiving the same closes.

## Fix
Module-level `threading.Lock` in `_fetch_yfinance` serializing every `yf.download` call, regardless of caller. Chain fetching stays parallel; only the OHLCV download is serialized.

## Repair
`scripts/repair_ohlcv_misalignment.py` re-fetches every symbol with bars ingested on the two corruption dates through the (now-serialized) single-symbol path — idempotent `ON CONFLICT DO UPDATE`, non-positive-close guard, `--dry-run` supported. Run once against prod after merge (or from a local checkout via the :5433 proxy).

## Deploy urgency
The deployed prod `quantcore-api` image predates this fix — **any Refresh All click in QuantUI re-corrupts prod OHLCV** until this ships (`deploy.yml` → test, then `prod-rollout.yml`). Re-run the repair after the fixed image is live if the button may have been used in between.

### DIFF
--- a/quantcore/repositories/ohlcv_repository.py
+++ b/quantcore/repositories/ohlcv_repository.py
@@ -15,6 +15,7 @@ from __future__ import annotations
 import datetime
 import enum
 import logging
+import threading
 from contextlib import closing
 from dataclasses import dataclass
 from typing import Optional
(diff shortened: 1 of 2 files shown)


thomasdfowler (commented):
@superDB12 @mijregas @musicalmacdonald Could one of you please approve this? There's a bit of a nasty bug that is fixed in this PR. Also, **please do not run a Refresh All from the UI or the CLI as it will corrupt the DB again**

JohnFunkCode (approved):

PR #133, opened 2026-07-25
Sidekick model selector: implementation plan + WP1-WP9 (issue #124)

### LINKED ISSUE #124: Add a configurable value in the keystore UI that allows you to change between Anthropic models (Fable 5, Opus 4.8, and Sonnet 5) in the sidekick
Add a configurable value in the keystore UI that allows you to change between Anthropic models (Fable 5, Opus 4.8, and Sonnet 5) in the sidekick.  Then connect this new configurable value up to the Anthropic keyproxy so the appropriate model is used for queries.

More capable models cost more per query.  The user should be able to select the model they want to use.  Here is a summary of the models for a user to select:
- Claude Fable 5: Anthropic’s most capable general model. Best for long-horizon, multi-file agentic work. (Pricing: $10M input, $50M output tokens).
- Claude Opus 4.8: The heavy-lifting flagship. Ideal for complex architecture decisions, deep refactoring, and multi-step reasoning. (Pricing: $5M input, $25M output tokens).
- Claude Sonnet 5: The recommended daily driver and default model. Offers the best balance of speed and intelligence for 90% of coding tasks. (Pricing: $2M input, $[1](https://platform.claude.com/docs/en/about-claude/models/overview)0M output tokens).

As of 7/23/2026 recommend making **Sonnet 5** the default

This change will require changes to the UI, as well as the Anthropic BYOK key proxy layer and all the layers in-between.

JohnFunkCode:
## Summary
- Adds `docs/proposals/sidekick-model-selector-plan.md`, a detailed, Sonnet-5-optimized implementation plan for a user-selectable Sidekick chat model (issue #124), checked against `docs/proposals/architectural-standard-v2.md`.
- Implementation work packets (WP1–WP9) land as subsequent commits on this branch, one packet per commit, in order.

Closes #124 once all packets land and are verified.

## Plan highlights
- Model catalog (Sonnet 5 default, Opus 4.8, Fable 5) lives in a neutral `quantcore/chat_models.py` constants module, injected by the registry into `SettingsService` and `ChatService` — keeps the services layer acyclic.
- New `user_settings(owner, chat_model, updated_at)` table + `SettingsService` + `GET/PUT /api/settings`, one service call deep, Pydantic `response_model=`.
- Model resolution (requested → stored → default) happens server-side in `ChatService`; independently allow-listed again at the keyproxy for defense in depth.
- No changes to the BYOK envelope/crypto path — `model` is a plain, non-secret field.

## Test plan
- [ ] `coverage run -m unittest discover && coverage report` (backend, per packet)
- [ ] `cd frontend && npx vitest run --coverage` (frontend, per packet)
- [ ] Manual: Settings dropdown persists model; chat-header quick-switch stays in sync; a crafted disallowed model is rejected by the keyproxy with `400`
- [ ] OpenAPI surface diff (`deploy.yml`) reflects the two new `/api/settings` routes and `ChatRequest.model`


frontend/src/components/settings/ModelSection.tsx
```diff
@@ -0,0 +1,92 @@
...
+      })
+      .finally(() => {
+        if (!cancelled) setReady(true);
+      });
+    return () => {
+      cancelled = true;
+    };
+  }, []);
+
+  const handleChange = async (event: SelectChangeEvent) => {
```
JohnFunkCode, line 25: **Warning:** This component maintains a separate `chatModel` state and calls `putChatModel()` directly, while the live chat header uses `ChatContext.selectedModel`. Because `ChatProvider` remains mounted while navigating between routes, changing the model here updates the server but not the live chat state; returning to the chat can still display and send the previous model. Please route this through `useChat().setSelectedModel` or a shared settings store so both controls stay synchronized.

frontend/src/components/settings/ModelSection.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const id = event.target.value;
+    setChatModel(id);
+    setSaving(true);
+    try {
+      await putChatModel(id);
```
JohnFunkCode: **Warning:** The UI updates `chatModel` optimistically, but failures from `putChatModel(id)` are not caught or surfaced, so the selector can show a model that was not persisted. Please await the mutation and either commit state only on success or roll back and display an error. The initial `getSettings()` chain also needs a rejection handler to avoid an unhandled promise when loading fails.

JohnFunkCode (commented):
## Code Review Summary

I found two functional issues that should be addressed before merge:

1. The Settings-page model selector is not synchronized with the live chat selector because it maintains separate state from ChatContext.
2. Model-selection persistence failures are hidden, leaving the UI showing a model that may not have been saved.

The frontend checks and secret scan pass. The backend check is currently red due to an exhausted mock in `tests.test_options_orchestrators.TestBackfill.test_per_date_state_machine`; the failing code is outside this PR's changed files.

See the inline comments for details. GitHub does not allow the PR author to submit a formal “Request changes” review on their own PR, so this was submitted as a comment review.

JohnFunkCode:
## Hermes Agent PR Review

**Verdict: APPROVE**

No findings. The model allow-listing, per-user persistence, API wiring, frontend synchronization, and fallback behavior are covered by the implementation and accompanying tests.

### Verification
- Python compilation succeeded.
- 53 targeted backend tests passed.
- Frontend tests could not run because the review environment does not have `vitest` and `@vitejs/plugin-react` installed.

*Posted from the latest scheduled PR review.*

JohnFunkCode:
Follow-up fix in #136 — the selector shipped here was broken for two of the three models it offers.

The turn request carried `fallbacks`, which only `claude-fable-5` accepts. `claude-sonnet-5` (the new default) and `claude-opus-4-8` reject the **entire turn** with a 400:

```
'claude-opus-4-8' does not support the `fallbacks` parameter
```

They don't ignore the unknown parameter, so every sidekick turn on either non-fable model failed outright. The request shape had been validated against fable and generalized to the other two without rechecking each one — provider parameters are not uniformly tolerated, and that assumption is what made this a shipping bug rather than a review catch.

#136 removes `fallbacks` from both mirrors (`quantcore/gateways/anthropic_gateway.py` and `keyproxy/providers/anthropic.py`) and rewrites the guard: `tests/test_anthropic_gateway.py` now stubs the SDK and asserts on the kwargs actually sent, for every allow-listed model on both the in-process and BYOK paths. Any future parameter added without checking all three fails the test instead of production.

It also carries two related fixes found in the same pass: the shared `max_tokens` budget is raised 8192 → 16384 (thinking tokens come out of it, and thinking volume varies by model, so visible-output headroom differed per model), and `stop_reason: "max_tokens"` is now surfaced as `Done(truncated=True)` with a note in the rail instead of ending the stream as if the model had chosen to stop.

Issue #135, opened 2026-07-26
Render markdown in the Sidekick rail (tables, emphasis, code) instead of plain text


JohnFunkCode:
The Sidekick rail renders assistant prose as **plain text**, not markdown. `SegmentView` in `frontend/src/components/chat/ChatRail.tsx:40-46` renders text segments as:

```tsx
<Typography variant="body2" sx={{ whiteSpace: 'pre-wrap', wordBreak: 'break-word' }}>
  {segment.text}
</Typography>
```

No markdown library is installed anywhere in `frontend/`. The practical effect: if the model emits a markdown table, the pipes show up literally and the columns do not line up (the font is proportional); `**bold**` renders as literal asterisks; backticks render as backtick characters.

This currently costs us real capability. Ranking / screening / comparison answers over a set of symbols are a core Sidekick use case, and a table is the natural format for them — see #121. The system prompt has to actively forbid tables, pipes, headings, bold, italics, and backticks purely because the rail cannot render them, which pushes the model toward a weaker one-line-per-symbol format.

## Proposal

Add markdown rendering to the text segment branch only.

- `react-markdown` + `remark-gfm` (GFM gives us tables), rendering to React elements.
- **No `rehype-raw`** — no raw HTML passthrough (see below).
- A component map rendering markdown nodes as MUI elements, with a whitelist: anything not on it degrades to text.
- Tables wrapped in an `overflow-x: auto` container. `RAIL_WIDTH = 400`, so anything past ~3 columns overflows regardless of renderer; the existing fullscreen toggle becomes the real wide-table mode.
- Reverse the anti-markdown clauses in `SYSTEM_PROMPT` (`quantcore/services/chat.py`) and update the matching prompt test in `tests/test_chat_service.py` (`test_system_prompt_forbids_markdown_tables`). **These must land together** — a rail that renders markdown plus a prompt that forbids it is the same bug in the other direction.
- Vitest coverage for the new rendering per arch-v2 Rules 8–9.

## This does not break live components

[...]

Issue #121, opened 2026-07-23
Sidekick chat: system prompt conflicts with explicit multi-ticker ranked-list requests


JohnFunkCode:
# Sidekick chat: system prompt conflicts with explicit multi-ticker ranked-list requests

## Summary
When a user asks the Sidekick chat to analyze **multiple tickers and return a prioritized/ranked list** (e.g. "here are 23 symbols, rank them in order of bullish signals"), the model frequently gathers data for every symbol but then renders **one** ticker's component and stops with a polite closing offer — never producing the ranked list the user asked for.

This is **intermittent** (it worked in some test cycles over the past several days) and reproduces in **both test and prod** (same system prompt, same model default), which rules out a code regression or an environment difference.

## Root cause (confirmed from prod logs)
The `SYSTEM_PROMPT` in `quantcore/services/chat.py` (~lines 37–62) is tuned for a **single-ticker side-rail** interaction and directly contradicts a **bulk ranked-report** request:

- *"After discussing a ticker, prefer showing the relevant component so the user sees live data … never restate numbers the component will display."* → nudges toward one ticker + one `show_component`.
- *"Be concise; this is a side rail, not a report."* → explicitly discourages the multi-row ranked report the user requested.

Given a very large tool-result context plus a "be concise, not a report" instruction, `claude-fable-5` resolves the tension by doing one deep dive and stopping. Because this is model judgment against a conflicting prompt (not deterministic code), it's intermittent and identical across environments.

### Evidence — live prod `quantcore-api` logs for the failing request
| Turn | Model action | Result size |
|------|--------------|-------------|
| 1 | 46 tool calls — `get_stock_price` ×23 + `get_technical_signals` ×23 | ~475,264 B of results returned to the model |
| 2 | **1** call — `show_component` (single ticker's `signals` panel) | 109 B |
| 3 | No tool calls → wrote a closing offer → `stop_reason: end_turn` | — |

No "degraded results" warning and no "iteration limit" error were logged — the model **voluntarily** stopped after rendering one component. So this is not truncation, not the `max_iterations` cap, and not related to the recently-reverted keyproxy PRs (#117/#118); the behavior is unchanged on the reverted test build.

## Proposed direction (for team discussion — NOT yet implemented)
Add an explicit carve-out to `SYSTEM_PROMPT` so an intentional multi-ticker request overrides the side-rail conciseness, roughly:

[...]

PR #136, opened 2026-07-26
fix(sidekick): drop fable-only `fallbacks`, surface truncated turns, stop shortening ranked lists

JohnFunkCode:
Follow-up to #133 (sidekick model selector). Three fixes found while exercising the shipped selector, plus two commits that were already on the previous branch and had no PR of their own.

## 1. `fallbacks` broke every non-fable turn

The selector shipped with a request shape that had been validated against `claude-fable-5` and generalized to the other two models without rechecking. `fallbacks` is fable-only: `claude-sonnet-5` (the default) and `claude-opus-4-8` reject the **entire turn** with a 400 —

```
'claude-opus-4-8' does not support the `fallbacks` parameter
```

— rather than ignoring the unknown parameter. So picking either non-fable model made the sidekick fail outright.

Removed the constants and `fallback_kwargs()` from both mirrors (`quantcore/gateways/anthropic_gateway.py` and `keyproxy/providers/anthropic.py`), leaving only parameters every allow-listed model accepts.

The guard rewrite matters as much as the fix: `tests/test_anthropic_gateway.py` now stubs the `anthropic` module and asserts on **the kwargs actually handed to the SDK**, for every allowed model on both the in-process and BYOK paths. A newly added unvetted parameter fails there instead of 400ing in production — which is exactly how this bug got in.

## 2. Output budget raised 8192 → 16384

Thinking tokens share `max_tokens`, and thinking volume varies by model (`claude-sonnet-5` has been observed returning `usage.thinking_tokens: 0`), so the headroom left for *visible* output differed per model on identical requests. Contributing cause of the short answers in #121.

Raised in all three places. keyproxy's `StreamTurnRequest` default moves too, but it's only a floor for callers that omit the field — `KeyProxyChatClient` always sends its own, so future budget changes need no keyproxy redeploy.

## 3. Short answers and silent truncation (#121)

Two halves of the same complaint: a ranked-list request came back short, and there was no way to tell whether the model chose to stop or ran out of room.

**Prompt** — `SYSTEM_PROMPT` said "Be concise; this is a side rail, not a report", which was being read as license to shorten a list the user explicitly asked for. Now:

- Concision is scoped to cutting padding, never content, with an explicit rule to cover every symbol named and to say which are missing and why when it can't.
- Ranking/scoring/comparing symbols is named as in scope and not personal investment advice, so those requests stop drawing disclaimers instead of answers.
- No web access is stated outright, plus a requirement to attribute any declined sub-request to the system prompt **by name** — a restriction disguised as a tool failure sends the user debugging the wrong thing.
- Markdown syntax is banned explicitly: the rail renders text plainly today, so asterisks and backticks show up literally. (Proposed markdown rendering is tracked separately in #135.) Multi-symbol answers get a one-line-per-symbol format instead of a table.
- Components are single-ticker, so multi-ticker questions are answered in prose rather than by rendering one and stopping.

**Truncation** — `stop_reason == "max_tokens"` now logs at WARNING and emits `Done(truncated=True)` instead of an ordinary `Done`. It stops even when the turn carries `tool_use` blocks: a block cut off mid-stream can hold incomplete argument JSON, and echoing that assistant turn back to the provider is itself an `invalid_request_error` risk — a partial answer the user can see beats a turn that dies opaquely one iteration later.

The flag reaches the wire for free (`api/sse.py` `asdict`s the event), `ChatContext` promotes it to message level, and `ChatRail` renders a muted italic caption. Without it a truncated ranked list is indistinguishable from a deliberately short one.

## Also included

Two commits from the previous branch that never got a PR: the backfill-orchestrator test date freeze (weekend flake) and the regenerated `openapi-surface.txt` for the `/api/settings` routes.

## Testing

| Suite | Result |
|---|---|
| `python -m unittest discover -s tests -t .` | 832 passed, 5 skipped |
| `npx vitest run --coverage` | 281 passed / 51 files, thresholds met |

New coverage: 3 backend truncation cases (flag set with partial text preserved; clean turn unmarked; truncated tool turn stops without executing the tool or making a second round trip), 3 vitest cases (caption shown, absent on a clean turn, scoped to the truncated message only), 5 request-shape cases across both provider paths, and the 6 prompt regression tests.

### DIFF
--- a/frontend/src/chat/ChatContext.tsx
+++ b/frontend/src/chat/ChatContext.tsx
@@ -297,6 +297,19 @@ export function ChatProvider({ children }: { children: ReactNode }) {
         });
       };
 
+      // A `done` frame carrying truncated=true means the answer was cut off
+      // at the output-budget ceiling. It's message-level, not another
+      // segment: it describes how the turn ended, and always lands last.
+      const markTruncated = () => {
+        setMessages((prev) => {
+          const next = [...prev];
+          const last = next[next.length - 1];
+          if (last?.role !== 'assistant') return prev;
+          next[next.length - 1] = { ...last, truncated: true };
+          return next;
+        });
+      };
+
       // If the stream dies without a terminal frame (server restart, network
       // drop), tool chips stuck in 'running' would spin forever — flip any
       // survivors to 'error' once the stream is over, however it ended.
@@ -338,7 +351,9 @@ export function ChatProvider({ children }: { children: ReactNode }) {
               setError(event.message);
             } else if (event.type === 'parse_error') {
               setError('Received an unreadable chunk from the server.');
-            } else if (event.type !== 'done') {
+            } else if (event.type === 'done') {
+              if (event.truncated) markTruncated();
+            } else {
               applyToAssistant(event);
             }
           },
--- a/frontend/src/chat/types.ts
+++ b/frontend/src/chat/types.ts
@@ -32,7 +32,9 @@ export type ChatStreamEvent =
     }
   | { type: 'directive'; directive: ChatDirective }
   | { type: 'error'; message: string }
-  | { type: 'done'; stop_reason: string }
+  /** `truncated` = the model hit its output budget mid-answer (stop_reason
+   *  "max_tokens"), not a clean finish. The prose above it is still valid. */
+  | { type: 'done'; stop_reason: string; truncated?: boolean }
   /** Client-side only: a frame whose data payload failed to JSON-parse. */
   | { type: 'parse_error'; raw: string };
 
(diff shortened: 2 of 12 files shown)


JohnFunkCode:
## Hermes Agent PR Review

**Verdict: APPROVE**

No bugs, security issues, performance problems, or actionable code-quality issues found.

### Verification results
- GitHub Actions: passed
  - Tests + wrapper smoke + OpenAPI surface
  - Frontend tests + typecheck + coverage
  - Secret scan
- Targeted Python tests: 70 passed, 1 failed because the review environment lacks the unrelated `pandas` dependency required by `test_registry_module_still_imports_lazily`.
- Frontend tests were not run locally because `frontend/node_modules` is absent.
- The PR’s added regression coverage appropriately checks both Anthropic request paths, all allow-listed models, truncation behavior, tool-use cutoff handling, and frontend rendering.

*Posted from the latest scheduled PR review.*

PR #141, opened 2026-07-26
test(frontend): statements 83.6→86.6%, functions 79.1→85.7%

thomasdfowler:
## Result

| Metric | Before | After | New floor |
|---|---|---|---|
| Statements | 83.62% | **86.62%** | 85 |
| Functions | 79.08% | **85.69%** | 84 |
| Lines | 85.69% | **88.37%** | 87 |
| Branches | 65.87% | 69.43% | 68 |

**281 → 331 tests**, typecheck clean.

Note that **lines were already above 85%** (85.69%) and never regressed — the metrics actually sitting under the bar were statements and functions, so those are what this targets. Branches improved as a side effect but remain below 85 by design; closing that gap is a much larger campaign and was deliberately scoped out.

## Approach

Rather than chase the percentage, the new tests drive behaviour that had **no** coverage: the plan/rung mutation dialogs and the securities toolbar were being *rendered* by existing tests but never *exercised*.

| Suite | Tests | What's now covered |
|---|---|---|
| `PlansPage` | 2 → 10 | grid `valueFormatter`s, row-click navigation, status filter refetch, archive confirm/cancel, Archive hidden for non-ACTIVE plans |
| `PlanDetailPage` | 2 → 10 | all ten summary fields, notes present/absent, archive → navigate back |
| `CreatePlanDialog` | 2 → 8 | symbol normalisation and advanced-param parsing asserted **on the request body**, error path, disabled-until-valid |
| `ExecuteRungDialog` | 1 → 7 | prefill from plan, edited submission, the omit-empty contract |
| `SecuritiesPage` | 5 → 28 | screener presets, sentiment dashboard, snapshot refresh (both variants + failure), tag persistence, sort model |

## Two contracts that were untested and easy to regress

1. **`ExecuteRungDialog` omits empty optional fields** — `tax_paid` and `notes` are dropped from the payload when blank rather than sent as `0` / `''`. Nothing asserted this; a refactor to "just send the state" would have silently started writing zero-tax rows.
2. **Filters survive corrupt `localStorage`** — both the tag filter and the sort model parse persisted JSON in a `try/catch`. There was no test that a bad value doesn't throw on mount.

## One thing to know when reading the diff

Assertions in `SecuritiesPage.test.tsx` are scoped to either the DataGrid or the sentiment table, never the whole document:

```ts
function inGrid() { return within(screen.getByRole('grid')); }
```

MUI's `Collapse` keeps the *collapsed* sentiment table mounted, so a bare `getByText('INTC')` matches two nodes and fails. Worth knowing before adding tests to that file.

## Verification

```bash
cd frontend
npx tsc -b --noEmit        # clean
npx vitest run --coverage  # 331 passed, thresholds enforced
```

Thresholds ratcheted `84/82/77/64` → `87/85/84/68`, pinned ~1pt under the measured baseline per the file's existing policy.

### Unrelated gotcha for local runs

The suite fails on **Node 25** with `TypeError: localStorage.clear is not a function`; CI pins **Node 24** and is green. Not addressed here — but a `.nvmrc` or `engines` field would stop the next person losing time to it.

### DIFF
--- a/frontend/src/components/plans/CreatePlanDialog.test.tsx
+++ b/frontend/src/components/plans/CreatePlanDialog.test.tsx
@@ -1,9 +1,23 @@
 import { afterEach, describe, expect, it, vi } from 'vitest';
 import { fireEvent, screen, waitFor } from '@testing-library/react';
 import CreatePlanDialog from './CreatePlanDialog';
-import { mockApi, renderWithProviders } from '../../testUtils';
+import { mockApi, renderWithProviders, type MockedApi } from '../../testUtils';
 
 afterEach(() => vi.unstubAllGlobals());
+
+/** The dialog's text inputs, in render order: symbol, then template name. */
+function textInputs() {
+  return Array.from(document.querySelectorAll('input[type="text"]')) as HTMLInputElement[];
+}
+
+function numberInputs() {
+  return Array.from(document.querySelectorAll('input[type="number"]')) as HTMLInputElement[];
+}
+
+function postBody(api: MockedApi) {
+  const call = api.calls.find(([url, init]) => url.includes('/api/plans') && init?.method === 'POST');
+  return call ? JSON.parse(String(call[1]?.body)) : null;
+}
 
 describe('CreatePlanDialog', () => {
   it('is hidden when closed', () => {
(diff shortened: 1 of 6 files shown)


JohnFunkCode (approved):
## Hermes Agent PR Review

**Reviewed head:** `b1987138d32a8ca4bc920992973753958ae5c8f2`
**Verdict:** APPROVE

**Associated issue:** None found. GitHub `closingIssuesReferences` is empty, and the PR title/body contains no unambiguous repository issue reference.

- **Fully met:** N/A — no associated issue.
- **Partially met:** None.
- **Unmet:** None.
- **Unverifiable:** None.

**Code findings:** None. The PR is limited to frontend test coverage additions and coverage-threshold ratcheting; the changed tests exercise the described behavior and the threshold values match the reported measured coverage.

**Verification:**
- `npx tsc -b --noEmit`: passed.
- `npx vitest run --coverage`: passed — 51 test files, 331 tests; Statements 86.62%, Branches 69.43%, Functions 85.69%, Lines 88.37%.
- Existing test-run stderr included React `act(...)` warnings and an HTML nesting warning; no test failures resulted.
- GitHub checks at review time: secret scan passed; frontend and backend gate checks were still pending.

PR #143, opened 2026-07-27
feat(sidekick): arbitrage tools + GenUI card

thomasdfowler:
The scanner shipped MCP-first, so the sidekick had **no access to it at all** — `ChatService` is constructed with `prices`/`fundamentals`/`sentiment`/`options` and nothing else. This adds three tools and a card.

| Tool | Purpose |
|---|---|
| `analyze_arbitrage_pair` | Full workup; `underlying` escape hatch for ad-hoc pairs |
| `scan_arbitrage` | Ranked summary rows |
| `list_arbitrage_universe` | What's curated, with staleness flags |

Plus `arbitrage_pair`, a GenUI card the model can render inline.

## Guarding the misread that motivates the whole feature

`analyze_pair` returns **both** `premium_discount_pct` (net of senior claims, −16.9% for MSTR) and `gross_premium_discount_pct` (−41.8%). An LLM glancing at that payload could confidently report a 42% discount to bitcoin — precisely the error the scanner was built to prevent. Four defences, because one wasn't enough:

1. **The tool description** names the right field and warns about the wrong one — and `test_arbitrage_tool_steers_the_model_to_the_net_discount` asserts that warning is still present, so a future trim fails CI rather than silently regressing.
2. **The system prompt** carries the same instruction.
3. **`scan_summary` omits the gross figure entirely** rather than trimming to it. A summary row is exactly where a misread happens, so the number simply isn't there.
4. **The card** puts the net figure in the headline slot and renders gross struck-through and labelled *"overstates what a shareholder gets"*, with a test pinning which number occupies the headline.

## Why `scan_arbitrage` doesn't call `scan`

A full scan returns ten candidates each carrying complete spread statistics, a NAV block, factors, reasons and `breaks_on`. That's far more than the model needs to rank them, and enough to crowd a conversation's context window. New `ArbitrageService.scan_summary` keeps the fields carrying the judgement — score, verdict, net discount, convergence, hedgeability, what breaks it — and drops the rest. The model calls `analyze_arbitrage_pair` for detail on anything interesting.

## GenUI compliance (arch-v2 Rules 8–9)

- Self-fetching from a ticker, scalar props, display-only
- Registered with matching strict specs in **both** `chat_tools.py` and `componentRegistry.tsx`
- **Formats but never derives** — every number comes from `quantcore/analytics` via the service
- **Curated pairs only.** `underlying` is deliberately *not* a prop: every registered prop is required, and there's no way to express an optional one without weakening the validator. Both registries reject it, asserted on both sides.
- Ships with vitest tests (loading/error/success + key values) and registry parity cases

## One robustness detail

The arbitrage handlers register **only when the service is injected**. A `ChatService` built without it hits the loop's existing unknown-tool path — a recoverable `is_error` the model can work around — instead of raising `AttributeError` on a `None` service mid-stream. Tested.

## Verification

| | Before | After |
|---|---|---|
| Backend tests | 969 | **986** |
| Frontend tests | 331 | **344** |
| Backend coverage | 86.2% | 86.2% |
| Frontend statements | 86.62% | **86.68%** |
| Frontend functions | 85.69% | **85.85%** |

Typecheck clean. Architecture guards pass. OpenAPI surface unchanged — no new routes; the card reuses `GET /api/arbitrage/pairs/{security}`.

**Note:** this needs a `prod-rollout.yml` dispatch to reach the deployed sidekick, same as the scanner itself did.

### DIFF
--- a/quantcore/services/chat.py
+++ b/quantcore/services/chat.py
@@ -358,11 +358,12 @@ class ChatService:
                 # Summary rows, not the full scan — ten candidates with complete
                 # statistics each would crowd the conversation's context window.
                 "scan_arbitrage": (
-                    lambda kinds=None, top_n=10:
+                    lambda kinds=None, top_n=10, days=365:
                     self._arbitrage.scan_summary(
                         kinds=[k.strip() for k in (kinds or "").split(",") if k.strip()]
                         or None,
                         top_n=top_n,
+                        days=days,
                     )
                 ),
                 "list_arbitrage_universe": lambda: self._arbitrage.get_universe(),
--- a/quantcore/services/chat.py
+++ b/quantcore/services/chat.py
@@ -73,6 +73,15 @@ the selected element directly; never echo the raw JSON back.
 Numbers you state in prose must come from tool results in this conversation,
 never from memory.
 
+On arbitrage results: quote the NET discount
+(nav.premium_discount_pct), never nav.gross_premium_discount_pct — the gross
+figure ignores debt and preferred stock ranking ahead of the common and
+routinely overstates the gap by 20+ points. If you cite both, label which one
+a shareholder actually gets. A "reject" verdict is a finding worth reporting
+plainly, not a failed lookup: say what the gap is and what stops it closing
+(no forcing mechanism, negative carry, an unhedgeable leg), rather than
+implying the tool found nothing.
+
 The tools above are your only source of market data. You have no web access:
 never attempt a web search, never fetch a URL, and never pull prices, news,
 filings, earnings figures, or any other market data from outside those tool
@@ -288,6 +297,7 @@ class ChatService:
         fundamentals,
         sentiment,
         options,
+        arbitrage=None,
         model: str = "claude-sonnet-5",
         effort: str = "medium",
         max_iterations: int = 8,
@@ -299,6 +309,7 @@ class ChatService:
         self._fundamentals = fundamentals
         self._sentiment = sentiment
         self._options = options
+        self._arbitrage = arbitrage
         self._model = model
         self._effort = effort
         self._max_iterations = max_iterations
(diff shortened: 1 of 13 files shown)


quantcore/services/chat.py
```diff
@@ -333,6 +344,29 @@ def __init__(
...
+                "analyze_arbitrage_pair": (
+                    lambda security, underlying=None, days=365:
+                    self._arbitrage.analyze_pair(
+                        security, underlying=underlying, days=days
+                    )
+                ),
+                # Summary rows, not the full scan — ten candidates with complete
+                # statistics each would crowd the conversation's context window.
+                "scan_arbitrage": (
+                    lambda kinds=None, top_n=10:
```
JohnFunkCode: **Warning:** `scan_arbitrage` does not accept or forward the documented `days` argument, so requests for a longer/shorter history silently use the fixed 365-day default. Add `days=365` to this handler, pass it to `scan_summary`, and expose it in the tool schema.

JohnFunkCode (changes requested):
## Hermes Agent PR Review

Reviewed head: c5df10f5dcf24cfdae0a1505ba06abec74dfa378
Verdict: REQUEST_CHANGES

### Code findings
- **quantcore/services/chat.py:361** — **Warning** — `scan_arbitrage` does not accept or forward the documented `days` argument, so requests such as “scan over two years” silently run with the fixed 365-day default despite the service and usage guide supporting 30–3650 days.
  **Fix:** Add `days=365` to the handler signature, include it in the `scan_summary(..., days=days)` call, and expose the corresponding integer property in the `scan_arbitrage` tool schema (with the documented bounds).

No associated issue found.

### Verification results
- Backend focused tests: 131 passed (`tests/test_arbitrage_service.py`, `tests/test_chat_protocol.py`, `tests/test_chat_service.py`).
- Frontend build: passed (`npm run build`).
- Frontend full suite: 343 passed, 1 failed (`src/components/settings/RotateKeyDialog.test.tsx`, unrelated to this PR’s changed files); the new `ArbitrageCard` tests passed (12 tests) and component-registry tests passed (21 tests).
- `git diff origin/main...pr-143 --check`: passed.

REQUEST_CHANGES

JohnFunkCode (approved):
## Hermes Agent PR Review

Reviewed head: daa7eaa691a223de9328199d29397defa0221741
Verdict: APPROVE

### Code findings
No blocking or non-blocking code findings. The prior finding about `scan_arbitrage` not forwarding `days` is fixed in this head; the handler, schema bounds, and regression test now cover custom windows.

No associated issue found. GitHub `closingIssuesReferences` is empty, and the title/body contain no unambiguous in-repository issue reference.

### Verification results
- `git diff origin/main...HEAD --check`: passed.
- GitHub Actions for this head: frontend tests/typecheck/coverage passed; backend tests, wrapper smoke, and OpenAPI surface passed; secret scan passed. Deployment jobs were skipped because deployment credentials were not configured.
- Local focused backend tests could not run because `pytest` is not installed in the review environment.
- Local frontend focused tests/build could not run because the checkout has no installed npm dependencies; GitHub Actions supplied the successful verification for this head.

APPROVE

PR #144, opened 2026-07-27
feat(arbitrage): visualization suite — spread, premium, scan, discovery

thomasdfowler:
Four visualization gaps, all closed, on **both** surfaces: each visual is a GenUI-compliant component the sidekick can render, and the new `/arbitrage` page composes the *same* components so the two can't drift.

## 1. The spread was invisible

`pairs.py` computes the residual series and `analyze_pair` **deliberately discards it** — a test pins that it never leaks into a payload, because a thousand floats is noise to an LLM. But it's the object every arbitrage number describes: a z-score of −1.35 tells you the gap is stretched; only the chart tells you whether it's been oscillating or walking away.

New `GET /pairs/{security}/spread-history` returns points plus mean, ±1σ/±2σ levels and the fitted trend line — **all precomputed**, so the chart draws without deriving (Rule 8). `spread_trend` now also returns `intercept`, letting the trend line be drawn without refitting in the browser.

## 2. Premium history is an approximation, and says so

Only the current capital structure is known, so every point applies *today's* holdings and senior claims to that date's prices. A reader taking that for real history would draw the wrong conclusion about how the gap behaved. So:

- payload carries `method: "current_capital_structure"` + an explicit note
- the chart renders the caveat inline, always
- tests assert both

It converges on truth as `arb_nav_snapshots` accumulates daily rows. Non-NAV securities get a structured refusal naming their family (*"GLD is a commodity_etf, not a NAV vehicle"*), which the card surfaces instead of a generic error.

## 3. Discovery needed its failures

A pass/fail list reported **"0 found"** for the obvious miner names and hid why. `discover_pairs(include_all=True)` now returns every tested pair with `passed` and `failed_because`, plotted against the critical-value lines — filled passed, hollow rejected.

Run live against NEM, it immediately shows **two different near-misses that the old output flattened into a single zero**:

| Pair | Correlation | Statistic | Rejected because |
|---|---|---|---|
| NEM ~ GC=F | **0.73** | −2.88 (vs −3.04) | not cointegrated |
| NEM ~ HG=F | 0.35 | **−3.80** (clears 5%) | correlation below the 0.4 floor |

The second is strongly cointegrated and thrown out purely by the correlation gate — visible for the first time.

## 4. Scan table + `/arbitrage` page

Route, nav entry, ranked table (score, verdict, net discount, convergence, hedge, what-breaks-it). Row click opens the pair detail composing the existing card + spread chart + premium chart for NAV vehicles. Collapsible discovery panel.

## Fixed while wiring the registry

`show_component` hard-required `ticker` — which would have **contradicted the registry** for `arbitrage_scan` (no props) and `arbitrage_discovery` (`symbols`), setting the model up to send a prop the validator then rejects. JSON Schema can't express per-component requirements without `oneOf`, so the required list is gone and the registry stays the single enforcement point. Two new tests: every registry prop must be declared in the schema, and no schema-level `required` list may return.

## Verification

| | Before | After |
|---|---|---|
| Backend tests | 987 | **1018** |
| Frontend tests | 344 | **382** |
| Statements | 86.68% | **87.35%** |
| Functions | 85.85% | **86.82%** |
| Branches | 69.67% | **70.51%** |
| Lines | 88.49% | **89.24%** |

Every frontend metric rose; floors ratcheted 88/86/85/69. Backend coverage 86.2%, typecheck clean, architecture guards pass, OpenAPI regenerated (99 routes).

Live-checked against the local API on the test DB: bands equal mean±σ exactly, trend intercept present, premium method label present, GLD correctly refused, and the NEM sweep returns all six tested rows with reasons.

**Note:** needs a `prod-rollout.yml` dispatch after merge, as usual.

### DIFF
--- a/api/routers/arbitrage.py
+++ b/api/routers/arbitrage.py
@@ -75,6 +75,11 @@ def discover(
     days: int = Query(365, ge=MIN_DAYS, le=MAX_DAYS),
     min_abs_correlation: float = Query(0.4, ge=0.0, le=1.0),
     require_economic_link: bool = True,
+    include_all: bool = Query(
+        False,
+        description="Also return every tested pair with a `passed` flag, so "
+                    "near-misses are visible (the discovery scatter needs this)",
+    ),
 ) -> QuantCoreJSONResponse:
     symbol_list = _split(symbols)
     if not symbol_list:
@@ -99,7 +104,31 @@ def discover(
             days=days,
             min_abs_correlation=min_abs_correlation,
             require_economic_link=require_economic_link,
+            include_all=include_all,
         )
+    )
+
+
+@router.get("/pairs/{security}/spread-history")
+def spread_history(
+    security: str,
+    underlying: Optional[str] = Query(None, max_length=32),
+    days: int = Query(365, ge=MIN_DAYS, le=MAX_DAYS),
+) -> QuantCoreJSONResponse:
+    return QuantCoreJSONResponse(
+        services().arbitrage.get_spread_history(
+            security.upper(), underlying=underlying, days=days
+        )
+    )
+
+
+@router.get("/pairs/{security}/premium-history")
+def premium_history(
+    security: str,
+    days: int = Query(365, ge=MIN_DAYS, le=MAX_DAYS),
+) -> QuantCoreJSONResponse:
+    return QuantCoreJSONResponse(
+        services().arbitrage.get_premium_history(security.upper(), days=days)
     )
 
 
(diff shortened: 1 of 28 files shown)


JohnFunkCode (approved):
## Hermes Agent PR Review

**Reviewed head:** `45b9214ff16a730750583f02a2575376ff3f6d8d`
**Verdict:** APPROVE

### Code findings
No Critical, Warning, or Suggestion findings. The new API routes, backend history/discovery calculations, frontend charts, registry entries, and associated tests were reviewed; no actionable correctness, security, performance, or error-handling issue was identified.

No associated issue found. GitHub `closingIssuesReferences` is empty, and the PR title/body contains no unambiguous in-repository issue reference.

### Verification
- Backend targeted tests: **126 passed** (`tests/test_arbitrage_nav.py`, `tests/test_arbitrage_service.py`, `tests/test_chat_protocol.py`) with a non-production SQLite DSN.
- Frontend changed-area tests: **59 passed** across the new arbitrage components, chart cards, and component registry.
- Frontend production build/typecheck: **passed** (`npm run build`).
- Full frontend suite: **381 passed, 1 failed**. The failure is `src/components/settings/RotateKeyDialog.test.tsx` and is outside this PR's changed files; it does not exercise the PR changes.
- API test module collection could not run in this isolated checkout because the repository requires a configured test database DSN for API tests; the targeted service/analytics/protocol tests ran successfully with the isolated SQLite DSN.

---
Reviewed automatically by Hermes Agent.

PR #149, opened 2026-07-27
PR 1: QuantUI rebrand, /harvester move, Portfolio route shell

JohnFunkCode:
## Summary
- PR 1 of the portfolio-lots plan ([docs/proposals/portfolio-lots-plan.md](docs/proposals/portfolio-lots-plan.md)) for #126.
- Rebrands the app to **QuantUI** (header, page title).
- Moves the harvester dashboard from `/` to `/harvester`, heading updated to "Harvester Dashboard".
- Adds a placeholder **Portfolio** page at `/` (nav positioned first) — intentionally no data yet, since making `/` a live portfolio view before per-owner scoping (PR 2) would leak one user's holdings to every logged-in user.

## Test plan
- [x] `cd frontend && npx vitest run --coverage` — 383/383 passing, coverage above ratchet floor (89.25%/87.35%/86.83%/70.51% vs 88/86/85/69)
- [x] `cd frontend && npx tsc --noEmit` — clean
- [x] `grep -rn "Harvest Ladder" frontend/src frontend/index.html` — no matches
- [x] Manually verified in browser: `/` shows Portfolio placeholder with correct nav order, `/harvester` shows "Harvester Dashboard" with nav highlighting
- [x] `python -m unittest discover -s tests -t .` — 1018/1018 passing (no backend changes in this PR)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #126, opened 2026-07-23
Add user specific external portfolios to the system


JohnFunkCode:
Enhance the current portfolio capability and extend it to act as a system to track external portfolios (managed outside of this tool, such as in eTrade, Charles Schwab etc).   

Use the user's google id to only show the user their portfolio data.  However, it is not necessary secure the portfolio data at rest.  
  
The user should be able to create, update, and delete portfolio data via the front end user interface.  The portfolio data should include "lot" data for any given symbol.   When a user creates a new lots they should be able to enter the purchase date, purchase price, and quantity of shared purchased.  When a user closes out a lot they should be able to provide the sale date, selling price, quantity sold.   If the number of shares sold doesn't match the number of shares in a lot, a new lot should be created for the remaining shares maintaining the original purchase date, purchase price. 

The user's portfolio with it's current prices, and Pofits/Losses should be shown on the main dashboard page.  The fields displayed for each lot should include:  Symbol, Current Price, Price Paid, Number of Shares, gain/Loss %, Gain/Loss $, Days the lot has been held, and dollars earned per day.  It should also display Today's gain/loss, 5 day gain/loss, 30 day gain loss, 90 day gain/loss and 1 year gain loss.
The portfolio display should be sortable by any field.

Totals for the portfolio should also be shown in a  portfolio summary section above the details for the stocks in the portfolio.  The summary section should include
  Total Investment
  Total Current Value
  Total Gain/Loss $
  Total Gain/Loss %

Dollars Per Day Gain/Loss: $0.11 

Additionally if the user clicks on the symbol for any of the stocks in the portfolio they should be taken to the details page for that symbol.

Adding symbols and lots to a user's portfolio should also be surfaced through a rest API and an MCP interface.

JohnFunkCode:
## Design decisions — consolidated

Record of the design discussion that reconciled this issue against the current
codebase. 21 questions plus follow-ons, all resolved. **Two of the decisions
amend the requirements as originally written** — flagged as ⚠️ below.

---

### Identity & scoping

**1. Canonical owner is a short handle** (`john`, `thomas`, …), not an email.
A new `owner_identities` table maps identity → owner, so the ES256 UI path
(`sub` = IAP email) and the HS256 MCP path (`sub` = short handle) converge on
one owner value. No data migration needed — existing `positions.owner='john'`
rows stay valid. This later doubles as the grant mechanism for shared/household
portfolios.

[...]

PR #150, opened 2026-07-27
feat: owner-scope the API from the authenticated principal (issue #126, PR 2)

JohnFunkCode:
## Summary
Part of [issue #126](https://github.com/JohnFunkCode/StockPortfolioManager/issues/126)'s portfolio-lots plan (PR 2 of the [plan](docs/proposals/portfolio-lots-plan.md)). Follows PR #149 (QuantUI shell / Portfolio route).

- Adds an `owner_identities` mapping table (`db/migrations/V3__owner_identities.sql`, mirrored in `quantcore/db.py`'s schema DDL) plus `OwnerIdentityRepository` + `IdentityService` to resolve an authenticated identity (IAP email or MCP token `sub`) to its canonical short-handle owner.
- Adds a `require_owner` FastAPI dependency (`api/auth.py`) that every owner-scoped route now depends on instead of accepting an `?owner=` query param — there's no legitimate cross-user read path on this endpoint. An unmapped identity is **never auto-provisioned**: it logs a `SECURITY unmapped_principal` event and returns 403 `not_provisioned`, identically for ES256 UI tokens and HS256 MCP tokens.
- Frontend: a full-page `RestrictedAccess` screen (no header/nav/ChatRail) renders when a call reports `not_provisioned`, wired through a small pub-sub in `api/client.ts` so `App.tsx` can early-return before the route tree mounts. A provisioned owner with zero positions sees a friendly `EmptyPortfolio` state instead of a bare empty table.
- Ops: `scripts/grant_quantui_iap_access.sh` now writes the matching `owner_identities` row in the same run as each IAP grant, and fails loudly (naming which half succeeded) if the DB write doesn't land — so IAP access and the DB mapping can't drift apart, which is exactly the state that would land a legitimate user on the restricted screen.

## Test plan
- [x] `python -m unittest discover -s tests -t .` — 1025 tests, `OK (skipped=5)`
- [x] OpenAPI route surface diff — 99 routes, matched
- [x] `cd frontend && npx tsc --noEmit` — clean
- [x] `cd frontend && npx vitest run --coverage` — 391/391 passing, coverage above thresholds
- [x] `scripts/grant_quantui_iap_access.sh` DB-write logic verified directly against the test database (idempotent conflict path + fresh-insert path); the live `gcloud iap` IAM calls were intentionally not executed here
- [ ] After deploy to test: verify with two identities — one provisioned (sees their own data) and one deliberately unmapped (sees `RestrictedAccess`, security event appears in Cloud Run logs)

### DIFF
--- a/api/auth.py
+++ b/api/auth.py
@@ -48,15 +48,22 @@ as its HMAC secret) fails closed.
 
 from __future__ import annotations
 
+import logging
 import os
 from dataclasses import dataclass, field
+from datetime import datetime, timezone
 from typing import Any, Optional
 
 import jwt
 from fastapi import Depends, HTTPException, status
 from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
 
+from quantcore.services.identity import UnknownIdentityError
+from quantcore.services.registry import get_services
+
 _TRUTHY = {"1", "true", "yes", "on"}
+
+_security_logger = logging.getLogger("quantcore.security")
 
 # auto_error=False so a missing/blank header reaches our dependency (which then either
 # bypasses, when AUTH_DISABLED, or raises a uniform 401) rather than FastAPI's generic 403.
@@ -122,6 +129,11 @@ class Principal:
 
     @property
     def owner(self) -> str:
+        """The raw authenticated identity (IAP email or MCP token ``sub``).
+
+        This is *not* the canonical short-handle owner used in ``positions.owner``
+        etc. — that resolution happens in ``require_owner``, via ``IdentityService``.
+        """
         return self.subject or (self.email or "unknown")
 
     @classmethod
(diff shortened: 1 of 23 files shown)


JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** `94a1db67a6afadb2d4d767e52e386279caf738d4`
**Verdict:** REQUEST_CHANGES

### Code findings

- **scripts/grant_quantui_iap_access.sh:88-90** — **Critical** — The onboarding write uses `ON CONFLICT (identity) DO NOTHING`, so rerunning it for an identity that is already mapped to a different owner silently preserves the old owner; the newly granted account can then be authenticated as the wrong user's portfolio (or denied access to its intended one). **Fix:** update the existing mapping with the requested owner and notes, or fail explicitly when the existing owner differs, then verify the stored mapping before reporting success.
  ```sql
  ON CONFLICT (identity) DO NOTHING
  ```

- **quantcore/db.py:393-400** — **Warning** — `init_schema()` creates `owner_identities` but does not seed the mappings defined in `db/migrations/V3__owner_identities.sql`, leaving a freshly initialized database with every authenticated principal unmapped and returning `403 not_provisioned` until Flyway or manual provisioning is run. **Fix:** make the bootstrap/deployment path apply the required seed data (or enforce V3 as a deployment prerequisite) and test a blank, auth-enabled database.
  ```sql
  CREATE TABLE IF NOT EXISTS owner_identities (...);
  ```

### Associated issue: #126 — Add user specific external portfolios to the system

The PR explicitly references issue #126 and its design-decision comment. Relevant PR 2 objectives:

- **Fully met:** derive the owner from the authenticated principal and remove `?owner=` from the owner-scoped REST routes; reject unmapped identities with a server-side 403 and security log; provide the standalone restricted-access UI; scope the screener, delta exposure, sentiment, and options snapshot capture to the resolved owner; add the identity mapping repository/service and onboarding path.
- **Partially met:** atomic/provisioned onboarding is implemented as a same-run IAM grant plus DB write, but conflicting existing mappings are silently retained (finding above), and the fresh `init_schema()` bootstrap does not seed the documented mappings (finding above).
- **Unmet:** none identified.
- **Unverifiable:** post-deploy verification with provisioned and unmapped identities, including Cloud Run security-log observation, was not run; the PR test plan also marks this as pending deployment verification.

### Verification

- `python -m compileall -q api quantcore main.py` — passed locally.
- `bash -n scripts/grant_quantui_iap_access.sh` — passed locally.
- `git diff --check` — passed locally.
- `cd frontend && npm run build` — passed locally.
- `cd frontend && npm test` — passed: 61 test files, 391 tests.
- Backend unittest suite — could not run locally because `QUANTCORE_DB_DSN`/the test PostgreSQL DSN is not configured in this environment.
- GitHub checks at review time: frontend and backend checks pending; gitleaks passed.

REQUEST_CHANGES

JohnFunkCode:
The code review from Hermes resulted in changes!

Issue #34, opened 2026-04-28
Add new cron job to populate fundamentals cache for all symbols in the galaxy of stocks we're watching.


JohnFunkCode:
Add new cron job to populate fundamentals cache for all symbols in the galaxy of stocks we're watching.  For each stock check to see if we have fundamentals (if not get them), if we do get the next earning date from get_earnings_calendar and retrieve fresh data a few days after earnings.

PR #156, opened 2026-07-28
fix: stop schema DDL from deadlocking concurrent writers (#152)

### LINKED ISSUE #152: Intermittent psycopg2 DeadlockDetected in full test suite (symbols vs plan_instances/plan_rungs)
## Summary

Two consecutive full-suite runs of `.claude/with-test-db.sh .venv/bin/python -m unittest discover -s tests -t .` against the QuantCore Postgres **test** DB produced `psycopg2.errors.DeadlockDetected` errors:

- Run 1: 4 errors, all inside `tests/test_harvester_service.py`
- Run 2: 5 errors, all inside `tests/test_harvester_repository_db.py` (different tests than run 1)

Both runs show the deadlock as a classic **AB-BA cycle between two relations**, consistently the same pair of relation OIDs within a given run (e.g. `symbols` and `plan_instances`/`plan_rungs`). Running `python -m unittest tests.test_harvester_repository_db -v` in complete isolation passed cleanly — all 15 tests, 455s, zero deadlocks — which rules out a bug in `quantcore/repositories/harvester_repository.py`'s `build_plan()` itself.

## What this implies

Since the deadlock only shows up in the full-suite run and not in isolation, something else running earlier in the suite (alphabetically before `test_harvester_repository_db.py` / `test_harvester_service.py`) is implicated — most likely a connection left open with an uncommitted transaction still holding a lock on the shared `symbols` table (a dimension table touched by most test files).

## What's been ruled out so far

- Every `get_connection()` call site in both `tests/` and `quantcore/` (repositories, services, gateways) is already wrapped in `with closing(...)`, and every write path calls `.commit()` (or `.rollback()` on exception) before the block exits. No naive "forgot to close" bug exists anywhere in the codebase.
- `harvester_repository.py`'s `build_plan()` sequences four separate short-lived connections (symbol upsert → OHLCV upsert → template lookup → plan_instance insert), each committing and closing before the next opens — confirmed clean, matches its clean isolated run.
- No unwrapped `get_connection()` calls or raw `psycopg2.connect()` calls exist outside `quantcore/db.py` and one `scripts/` migration tool (not exercised by tests).
- `ThreadPoolExecutor` usage in `quantcore/services/prices.py:1842` and `quantcore/services/options.py:1064/1391` is always used as a context manager (blocks on exit until all workers finish), and doesn't touch `plan_instances`/`plan_rungs`, so it doesn't explain the harvester-specific table pair.
- Checked `pg_stat_activity` on the test DB post-hoc (read-only, via the project's existing venv + `with-test-db.sh` DSN): 0 lingering backends, no held locks. This doesn't rule root cause in or out — a `DeadlockDetected` error means Postgres itself already broke the cycle by aborting one transaction, so the offending backend cleans up and vanishes right after the error fires. A post-hoc check will always come back clean regardless of cause.

## Leading hypothesis (unconfirmed)

A genuine Postgres deadlock requires two backends *simultaneously* blocked on each other, which needs real concurrency. `unittest discover` runs single-threaded/sequential, so the leak/race must come from something that creates real concurrent DB sessions — e.g. a `TestClient`-driven FastAPI endpoint whose sync route handler runs in Starlette's thread pool while another connection is opened elsewhere, or a genuinely leaked/zombie connection from a previous *interrupted* run of the suite (Ctrl+C, timeout) that's still alive on the server holding locks from whatever it last touched, colliding with a fresh connection from the current run.

## Next step (not yet done)

Catch the deadlock live: kick off a fresh full-suite run in the background and poll `pg_stat_activity` / `pg_locks` (filtered to `wait_event_type = 'Lock'`) while it's mid-flight, to identify the actual two backends/queries forming the AB-BA pair. That will pinpoint the actual leaking code path instead of further static analysis.

---
*Filed from an investigation session; no code changes have been made yet.*

JohnFunkCode:
Fixes #152.

## The bug

`quantcore/db.py:init_schema()` ran all 65 DDL statements in a **single
transaction**, so it held locks on every one of the ~20 tables it had touched
until the final commit:

| stmt # | table |
|--------|-------|
| 1 | `symbols` |
| 2–4 | `ohlcv` |
| 34–36 | `plan_instances` |
| 37–38 | `plan_rungs` |

At statement #34 the DDL still holds `symbols` from statement #1.
`harvester_repository.build_plan()` takes the same two tables in the opposite
order — `RowExclusiveLock` on `plan_instances`/`plan_rungs`, then `symbols` for
its FK lookup:

```
DDL       : holds symbols        -> wants plan_instances
build_plan: holds plan_instances -> wants symbols        = AB-BA
```

`ShareLock` (`CREATE INDEX`) conflicts with `RowExclusiveLock` (`INSERT`), so
the cycle is real. This matches the reported signature exactly: the same pair of
relation OIDs within a run, `symbols` <-> `plan_instances`/`plan_rungs`.

**The window is ~3 seconds wide, measured** — 8 full DDL runs in the first 71
seconds of a suite run (2.24–2.99s each, 20.3s total), 65 statements each a
round trip through the Cloud SQL proxy. Live `pg_locks` capture during a run
showed a 13.4s sustained block between two connections of the same unittest
process:

```
WAITER  pid=2479068  CREATE INDEX IF NOT EXISTS idx_ohlcv_lookup ON ohlcv (...)
        granted=False  relation/ShareLock         rel=ohlcv
BLOCKER pid=2479064  INSERT INTO ohlcv (... 'ZZHARVREPO' ...)
        granted=True   relation/RowExclusiveLock  rel=ohlcv  xid=822070
```

### Why it only failed in the full suite

`init_schema()` had a once-per-process guard, but two callers bypassed it:

1. **`api/main.py`** — `create_app()` called `init_schema()` *unconditionally*,
   from **16 call sites** in the suite, most of them in `setUp`.
2. **`quantcore/repositories/options_repository.py`** — `OptionsStore(dsn=...)`
   never set the flag, since `target_dsn != DB_DSN`, so the DDL re-ran on every
   construction. Latent today, same bug.

`tests/test_arbitrage_repository.py` also re-ran it in `setUpClass`.

Running `test_harvester_repository_db` alone never constructs an app, so the DDL
runs once before any writes — no overlap, no deadlock. That's why isolation
"cleared" the repository: **`harvester_repository.py` is the victim, not the
cause, and is unchanged in this PR.**

## The fix

1. **Run the DDL in autocommit.** Every statement is individually idempotent
   (`CREATE ... IF NOT EXISTS`, `ADD COLUMN IF NOT EXISTS`), so per-statement
   commit converges to the same schema while releasing each lock immediately.
   With one lock held at a time a two-table cycle cannot form. Plus
   `SET lock_timeout = '10s'` so blocking behind a live writer fails loudly
   instead of hanging.
2. **Stop re-running the DDL at arbitrary times.** New `ensure_schema(dsn=None)`
   is keyed on a set of initialized DSNs (so a second database is initialized
   once too, not per construction). `create_app()`, `OptionsStore.__init__`,
   `get_connection()` and `test_arbitrage_repository.setUpClass` all go through
   it. `main.py` and `fastMCPTest/options_analysis.py` still call `init_schema()`
   directly — they're process entry points where an explicit one-shot init is
   the intent, and `init_schema` now records the DSN so `get_connection` won't
   repeat it.
3. **Regression guard** — `tests/test_schema_bootstrap.py` (6 tests, no database
   required, fake connection): the DDL runs in autocommit with a `lock_timeout`,
   executes every statement, runs at most once per process per DSN, and two
   `create_app()` calls trigger it at most once.

## Verification

`tests/test_schema_bootstrap.py`: 6/6 pass.

Two consecutive full-suite runs against the test DB, instrumented to count DDL
runs — results appended as a comment once they land.

| | before | after |
|---|---|---|
| full DDL runs | 8 in the first 71s | 1, at startup |
| `DeadlockDetected` | 4–5 per suite | 0 |

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -44,9 +44,9 @@ from .json_response import QuantCoreJSONResponse  # noqa: E402
 
 def create_app() -> FastAPI:
     """Application factory mirroring the Flask ``create_app`` contract."""
-    from quantcore.db import init_schema
+    from quantcore.db import ensure_schema
 
-    init_schema()
+    ensure_schema()
 
     app = FastAPI(
         title="QuantCore REST API",
--- a/quantcore/db.py
+++ b/quantcore/db.py
@@ -502,21 +502,45 @@ class _PGConn:
 
 # Schema is ensured once per process on first connection, so callers never
 # see a missing table regardless of which entry point started the process.
-# Entry points may still call init_schema() explicitly; it sets the flag too,
-# so the DDL is not re-run on the first connection afterwards. The lock keeps
-# concurrent first connections (threaded Flask / MCP servers) from running the
-# DDL in parallel — PostgreSQL's CREATE ... IF NOT EXISTS can fail under
-# concurrent execution.
-_schema_ready = False
-_schema_lock = threading.Lock()
+# Entry points should call ensure_schema() rather than init_schema(): it
+# records the DSN, so the DDL is not re-run on the first connection afterwards.
+# The lock keeps concurrent first connections (threaded Flask / MCP servers)
+# from running the DDL in parallel — PostgreSQL's CREATE ... IF NOT EXISTS can
+# fail under concurrent execution.
+#
+# Keyed by DSN rather than a single bool so that pointing a repository at a
+# second database (OptionsStore(dsn=...)) initializes it once too, instead of
+# on every construction.
+_schema_ready_dsns: set[str] = set()
+# Reentrant: ensure_schema() holds it across init_schema(), which records the
+# DSN under the same lock.
+_schema_lock = threading.RLock()
+
+
+def ensure_schema(dsn: str = None) -> None:
+    """Run the schema DDL at most once per process, per DSN.
+
+    Callers that just need the tables to exist should use this rather than
+    ``init_schema()``: the DDL takes ~3s against a managed instance and locks
+    ~20 tables while it runs, so re-running it on every application factory
+    call or repository construction is both slow and a source of lock
+    contention with concurrent writers.
+    """
+    target_dsn = dsn or DB_DSN
+    if target_dsn in _schema_ready_dsns:
+        return
+    with _schema_lock:
+        if target_dsn in _schema_ready_dsns:
+            return
+        init_schema(target_dsn)
+        # init_schema() records it too; recorded here as well so the
+        # once-per-process gate holds however init_schema is reached.
+        _schema_ready_dsns.add(target_dsn)
 
 
 def get_connection() -> _PGConn:
     """Get a connection to the QuantCore PostgreSQL database."""
-    if not _schema_ready:
-        with _schema_lock:
-            if not _schema_ready:
-                init_schema()
+    ensure_schema()
     return _PGConn(psycopg2.connect(DB_DSN))
 
 
@@ -526,16 +550,29 @@ def _split_schema(schema: str) -> list[str]:
 
 
 def init_schema(dsn: str = None) -> None:
-    """Initialize the database schema if tables don't exist."""
-    global _schema_ready
+    """Initialize the database schema if tables don't exist.
+
+    Each statement commits on its own. The DDL is idempotent
+    (``CREATE ... IF NOT EXISTS`` / ``ADD COLUMN IF NOT EXISTS``), so
+    per-statement commit converges to the same schema, and it keeps the DDL
+    from holding locks on every table it has touched until one final commit.
+    That accumulation is what let a concurrent writer form an AB-BA deadlock:
+    the DDL holds ``symbols`` (statement 1) while reaching for
+    ``plan_instances`` (statement 34), while a writer holds ``plan_instances``
+    and reaches for ``symbols`` through its foreign key.
+
+    ``lock_timeout`` turns a wait behind a live writer into a loud, diagnosable
+    error instead of an indefinite hang.
+    """
     target_dsn = dsn or DB_DSN
     conn = psycopg2.connect(target_dsn)
     try:
+        conn.set_session(autocommit=True)
         with conn.cursor() as cur:
+            cur.execute("SET lock_timeout = '10s'")
             for stmt in _split_schema(_SCHEMA):
                 cur.execute(stmt)
-        conn.commit()
     finally:
         conn.close()
-    if target_dsn == DB_DSN:
-        _schema_ready = True
+    with _schema_lock:
+        _schema_ready_dsns.add(target_dsn)
--- a/quantcore/repositories/options_repository.py
+++ b/quantcore/repositories/options_repository.py
@@ -30,7 +30,7 @@ from contextlib import closing
 from datetime import datetime, timezone
 from typing import Optional
 
-from quantcore.db import get_connection, init_schema
+from quantcore.db import get_connection, ensure_schema
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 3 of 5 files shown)


JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** `0d0a35f84e7d03e3632c99c21ca42fa5193a6dbd`
**Verdict:** APPROVE

### Code findings
No actionable bugs, security issues, performance regressions, or code-quality issues found in the five-file change. The schema initialization change uses a process-local reentrant lock, tracks initialized DSNs, enables autocommit for idempotent DDL, and applies a lock timeout. The added tests cover autocommit, statement execution, per-DSN gating, direct initialization bookkeeping, and repeated application-factory calls.

Associated issue: #152 — Intermittent psycopg2 DeadlockDetected in full test suite (symbols vs plan_instances/plan_rungs).

- **Fully met:** Prevent repeated schema bootstrap per process and DSN; release schema DDL locks per statement; add a bounded lock wait; add regression coverage for the bootstrap behavior.
- **Partially met:** None.
- **Unmet:** None.
- **Unverifiable:** The issue's claimed two consecutive full-suite post-fix runs are reported in the PR description but were not independently rerun in this environment because the repository's database/test-DB setup was unavailable.

### Verification
- `python3 -m compileall -q quantcore/db.py api/main.py quantcore/repositories/options_repository.py tests/test_schema_bootstrap.py tests/test_arbitrage_repository.py` — passed.
- `python -m unittest tests.test_schema_bootstrap -v` — five dependency-free tests passed; the sixth application-factory test could not run because the isolated environment lacked the declared `yfinance` dependency. This is an unavailable check, not a code failure.
- No database-backed or full-suite test was run because no test database/credentials were available.

*Formal APPROVE was attempted but GitHub rejected it because the authenticated account is the PR author; this top-level comment preserves the complete review and intended verdict.*

JohnFunkCode:
## Verification: two consecutive full-suite runs, both clean

```
run 1: Ran 1045 tests in 689.976s  OK (skipped=5)
run 2: Ran 1045 tests in 697.370s  OK (skipped=5)
```

**Zero `DeadlockDetected` in either run**, against a reported 4–5 per run before
the fix. No `lock_timeout` fired either, so nothing is merely being masked by the
new timeout — the DDL simply isn't colliding with writers any more.

Both runs were instrumented to trace every `init_schema()` call. Identical
results:

```
[+   3.0s] DDL run #1 in 2.41s   <- api/main.py:117  app = create_app()
[+ 681.0s] DDL run #2 in 0.00s   \
[+ 681.0s] DDL run #3 in 0.00s    > tests/test_schema_bootstrap.py (fake conn)
[+ 681.0s] DDL run #4 in 0.00s   /
```

**One real DDL run per suite**, at import time, before any test writes. The three
at +681s are the new regression tests driving a fake connection — 0.00s, no
database touched.

For contrast, the pre-fix baseline (salvaged from an instrumented run I killed
early) was **8 full DDL runs in the first 71 seconds alone**, 2.24–2.99s each,
20.3s of locks-held-on-~20-tables inside 71s of wall clock — and that is just the
opening minute of a ~690s suite.

A `pg_locks` poller ran across the whole window at 0.5s intervals. Contention
events cluster only in the pre-fix runs (09:31:37–09:31:50, a single 13.4s
sustained block, and 09:52:41–09:52:42 from two of my own diagnostic runs
overlapping). **Nothing after the fix landed.** An idle-in-transaction watcher
with a 5s threshold logged zero leaks throughout — consistent with the finding
that no connection was ever leaked; the transaction was simply far too wide.

PR #172, opened 2026-08-09
docs+feat: schema evidence layer for issue #165 (PR 1 of 4)

### LINKED ISSUE #165: init_schema() and Flyway both own the deployed schema — flyway info can't be trusted as evidence of prod state
## Summary

Two systems own the schema of deployed databases, and they don't know about each other:

- **Flyway** (`db/migrations/V*.sql`) — the declared source of truth since Phase 1.
- **`init_schema()`** (`quantcore/db.py:567`) — 22 tables of idempotent DDL, run via
  `ensure_schema()` on **every application startup** (`api/main.py:49`, `main.py:621`,
  `quantcore/repositories/options_repository.py:52`).

Because `init_schema()` is mirrored forward on every schema change (a standing project rule)
and runs before Flyway ever does, deployed databases arrive at the correct *shape* without
Flyway's participation. Flyway's ledger then records history that never happened.

## How it surfaced

During the #162 watchlist rollout (2026-07-28), `flyway info` against **prod** read:

```
| Versioned | 3 | owner identities  | SQL | 2026-07-27 10:03:03 | Success |
| Versioned | 4 | portfolio lots    | SQL |                     | Pending |
| Versioned | 5 | watchlist         | SQL |                     | Pending |
```

Two migrations pending, one of them (V4) unnoticed for over a week. But a read-only check of
the actual database showed **every V4 and V5 object already present**: `lot_sales`,
`watchlist`, and all nine new `positions` columns (`status`, `parent_lot_id`, `trade_date`,
`fees`, `acquisition_type`, `covered`, `fx_rate_at_purchase`, `basis_adjustment`,
`wash_sale_disallowed`).

`flyway migrate` then "applied 2 migrations" in 758ms, consisting almost entirely of:

```
WARNING: DB: column "status" of relation "positions" already exists, skipping
WARNING: DB: relation "lot_sales" already exists, skipping
WARNING: DB: relation "watchlist" already exists, skipping
```

Net effect: two ledger rows. No DDL.

## Why this is a problem

1. **`flyway info` is not evidence.** "Pending" did not mean the objects were missing, so the
   command can't answer the one question you run it for before a rollout. The inverse is worse:
   genuine drift — a hand-applied change, a migration that half-failed — would be equally
   invisible, because `init_schema()` quietly re-converges the shape on the next pod start.
2. **The migrate-before-deploy ordering was never load-bearing.** We sequenced the #162 rollout
   carefully so the `watchlist` table would exist before the new API image landed with no
   `watchlist.yaml` fallback (#83 decision 7). That ordering turned out not to matter — the
   table had been created by `init_schema()` long before. The safety came from a mechanism we
   weren't reasoning about, and the mechanism we *were* reasoning about did nothing. Right
   outcome, wrong model — and the model is what we'll rely on next time.
3. **Only shape converges, never data.** `init_schema()` creates tables; it cannot seed them or
   run a backfill. So a migration that is pure DDL is silently redundant, while one carrying an
   `UPDATE`/backfill is silently *required* — and nothing distinguishes the two at a glance.
   V4's three backfill `UPDATE`s ran against an empty prod `positions` table and were no-ops by
   luck of timing, not by design.
4. **Failure mode is deferred, not prevented.** If the two ever disagree, the loser is whichever
   process starts last, and the symptom appears at an arbitrary later pod restart rather than at
   deploy time.

## Options

Not proposing a specific fix — this wants a decision, not a patch:

- **Retire `init_schema()`'s DDL on deployed environments.** Flyway owns schema; `ensure_schema()`
  becomes a no-op (or a verify-only check) when a `flyway_schema_history` table is present. Keeps
  the local/test bootstrap path that tests depend on (`tests/test_schema_bootstrap.py` asserts
  `create_app()` calls it unconditionally — that contract would change).
- **Keep both, add a drift check.** `init_schema()` stays, but CI or startup compares the live
  schema against the migration-derived expectation and fails loudly on divergence. Cheaper, keeps
  the current bootstrap ergonomics, but keeps two writers.
- **Baseline honestly and move on.** Accept that deployed DBs are shape-converged by
  `init_schema()` and treat Flyway as a changelog for data/backfill migrations only. Weakest
  option, but worth naming so it's a choice rather than the status quo by default.

Whichever way it goes, `flyway info` should stop being quoted in runbooks as proof of prod state
until it actually is.

## Notes

- Discovered while rolling out #162; both prod migrations are now applied and prod is at v5 with
  227 watchlist rows, so **nothing is currently broken**. This is a correctness-of-process issue.
- There was no documented Flyway invocation at all until this rollout — `db/flyway.conf`
  deliberately holds no credentials and the commands were reconstructed from the DSN each time.
  Worth committing a `scripts/flyway.sh` wrapper alongside whatever is decided here.

JohnFunkCode:
## Summary

PR 1 ("Evidence") of [docs/proposals/schema-ownership-plan.md](docs/proposals/schema-ownership-plan.md), which fixes #165: `init_schema()` and Flyway are two uncoordinated owners of the deployed schema, and `flyway info` is not evidence of what a database actually contains. This PR adds read-only tooling to see whether the two agree — no enforcement/behavior change yet (that's PR 2/4).

## What landed

- **`db/baseline/V1__quantcore_baseline.sql`** — written-down baseline extracted from `init_schema()` as of `9a9d204^` (16 tables, 0 `ALTER TABLE`). Kept outside `flyway.locations` (Decision D3) so it doesn't run as a live migration.
- **`quantcore/schema_introspect.py`** — `describe_schema()` (introspects tables/columns/indexes/constraints via `information_schema`/`pg_indexes`/`pg_constraint`), `diff_schemas()` (MISSING/EXTRA/MISMATCH; D4: missing objects are errors, extra objects are warnings), and `scratch_database()`/`snapshot_from_dsn()` for building a snapshot from a live `init_schema()` run against a throwaway database. `tests/test_schema_introspect.py`: 13 DB-free unit tests.
- **`db/schema_snapshot.json`** — committed snapshot of what `init_schema()` actually produces today (22 tables), plus **`scripts/check_schema_snapshot.py`** (`--update` to regenerate, plain run to diff) wired into `.github/workflows/deploy.yml`'s `gate` job as a CI drift check.
- **`scripts/schema_check.py`** — read-only `--test`/`--prod` evidence command (mirrors `flyway.sh`'s DSN selection convention). Prints the live-schema-vs-snapshot diff and, separately, the Flyway changelog labelled for what it is. Opens no write transaction, takes no DDL lock, safe to run against prod any time.
- **Docs** — `readme.md`, `CLAUDE.md`, and `scripts/flyway.sh`'s header now point at `scripts/schema_check.py` instead of "check the objects directly"; the plan's own checkpoint log gained its PR 1 row.

## What it found

Ran `scripts/schema_check.py` against **both live databases**:

- test (`127.0.0.1:5434` via Cloud SQL proxy): `22 tables, 0 missing, 0 extra`, Flyway ledger `applied through V5__watchlist (2026-07-28)`
- prod (`127.0.0.1:5433` via Cloud SQL proxy): `22 tables, 0 missing, 0 extra`, same Flyway ledger line

**Both clean — no drift on either side.** `init_schema()` and Flyway agree today. This is the baseline PR 2's CI parity test will hold going forward.

## Verification

- `python -m unittest tests.test_schema_introspect -v` → 13/13 pass
- `check_schema_snapshot.py --update` → wrote 22-table snapshot; re-run clean; drift-detection proven by injecting a throwaway table (`EXTRA table zzz_drift`, exit 1), then reverting and re-confirming clean, exit 0
- `schema_check.py --test` and `--prod` → both clean, as above

Part of #165. Next up: PR 2 (`feat/schema-parity`) adds a CI test comparing `init_schema()` output against baseline+migrations output.


JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** `b76cd8187851200b3b73acec282e5e655111a2ec`
**Verdict:** REQUEST_CHANGES

### Code findings

- **.github/workflows/deploy.yml:115** — **Warning** — The new `python scripts/check_schema_snapshot.py` step uses `QUANTCORE_DB_DSN`, but this workflow sets that DSN at line 66 with password `***` while the Postgres service is configured with `POSTGRES_PASSWORD: quantcore` at lines 55–57, so the new database-backed check will fail authentication in CI. **Fix:** make the gate DSN password match the service password (or construct it from the same workflow variable) before invoking the snapshot check; do not commit a real credential.

### Associated issue and objectives

Associated issue: [#165](https://github.com/JohnFunkCode/StockPortfolioManager/issues/165) — *init_schema() and Flyway both own the deployed schema — flyway info can't be trusted as evidence of prod state*.

The PR's scoped PR-1 objectives are **Fully met**: it adds the baseline reference, schema introspection/diffing, a committed `init_schema()` snapshot, a CI drift check, a read-only live-schema evidence command, documentation updates, and unit tests. The broader issue objective is **Partially met**: the evidence/drift-comparison portion is delivered, but the two-owner deployment behavior remains for the later PRs described by the issue plan.

### Verification

- `python -m compileall -q quantcore/schema_introspect.py scripts/check_schema_snapshot.py scripts/schema_check.py tests/test_schema_introspect.py` — passed.
- `python -m unittest tests.test_schema_introspect -v` — unavailable in this review environment because `psycopg2` is not installed; CI installs `requirements-dev.txt`, which includes it. The test module imports `psycopg2` eagerly despite its DB-free test cases.
- Live Postgres/snapshot verification was not run in this environment.

PR #173, opened 2026-08-09
test: enforce that init_schema() and the Flyway files agree (#165, PR 2)

JohnFunkCode:
PR 2 (Enforcement) of [`docs/proposals/schema-ownership-plan.md`](docs/proposals/schema-ownership-plan.md), for issue #165. PR 1 (#172) built the evidence tooling; this makes divergence between the two owners of the schema impossible to merge.

## What lands

- **`tests/test_schema_parity.py`** — builds one scratch database from `init_schema()` and another from `db/baseline/V1__*.sql` + every `db/migrations/V*.sql` in parsed-integer version order, and asserts the two descriptions are identical. The failure message carries the full MISSING/EXTRA/MISMATCH list, because the diff *is* the bug report. `MigrationOrderTests` pins V10-after-V9 ordering and baseline-first contiguity; `SkipGuardTests` pins that the guard **fails rather than skips** when `CI` is set.
- **`db/migrations/V6__parity_backfill.sql`** — the first parity run's finding, resolved.
- Docs: `CLAUDE.md`, `AGENTS.md`, `readme.md` now state the three-file rule as CI-enforced rather than a convention, and the plan doc gains the "What the first parity run found" section plus the PR 2 checkpoint row.
- Two carry-overs from PR 1, kept here deliberately rather than split into their own PR: the coverage repair that lifts `quantcore/schema_introspect.py` from 52% → 98%, and the `sys.path` preamble both operator scripts shipped without.

## What the first parity run found

Three `MISSING table` lines — `arb_nav_snapshots`, `gex_history`, `user_settings` — and **zero MISMATCH lines**.

The plan predicted constraint-name drift, index-name drift, and a `positions.quantity` type mismatch. None of that exists: every table both sides have is identical down to columns, indexes, and constraints. The drift was purely *additive* — three tables added to `_SCHEMA` that never got a migration. So the interesting failure mode here is "someone added a table and forgot the migration", not "the two sources describe the same table differently".

Under decision **D6, prod reality wins**. Prod has all three (verified read-only against `quantcore-prod-20260606`: `schema_check.py --prod` reports `22 tables, 0 missing, 0 extra`, plus a direct `information_schema.tables` query naming them), so the migrations were the wrong side. V6 is idempotent DDL copied verbatim from `_SCHEMA`. **`V2`–`V5` untouched.**

V6 will be a no-op on every existing database — that is the point, not a sign it's unnecessary. Its job is to make the *file set* a complete description of the schema. The case it actually fixes is a database built from empty: a new environment, a scratch database, a restore drill.

The drift was invisible precisely because nothing broke: `init_schema()` runs on every application startup and created these tables before Flyway ever looked. No outage, no error, no log line — the only symptom was `flyway info` saying `V5` while the database had 22 tables. This is the concrete instance of the claim in `CLAUDE.md` that `flyway info` is a changelog view, not evidence.

## Verification

- Full suite: **1228 tests, OK, 5 skipped** (up from 1221 — exactly the 7 new tests; the same 5 pre-existing skips).
- `./.claude/with-test-db.sh .venv/bin/python -m unittest tests.test_schema_parity -v` → 7/7 pass in 10.4s against a real Postgres. The database-backed case genuinely runs; it does not skip.
- `python scripts/check_schema_snapshot.py` → `schema snapshot up to date (22 tables)`. `db/schema_snapshot.json` is unchanged, as expected — V6 adds nothing `init_schema()` didn't already have.
- No deployment needed. Nothing here is imported by any service, router, MCP wrapper, or `main.py`.

## Gotcha worth reading

Both PR 1 scripts shipped without the `sys.path.insert(REPO_ROOT)` preamble every other script in `scripts/` has, so `python scripts/schema_check.py --prod` from the repo root died with `ModuleNotFoundError: No module named 'quantcore'` before printing anything. CI sets `PYTHONPATH: .` job-wide, which is exactly why it passed review — the failure only reproduces interactively, and interactive is the only way `schema_check.py` is ever invoked. **A command whose entire purpose is hand-invocation cannot be validated solely by a CI job that pre-configures its environment.**

Related: PR 1's CI run never actually reached its new "Schema snapshot up to date" step — it aborted earlier at the PR-only `Diff coverage` gate, which the post-merge push run then skips. The step's first real execution was run `31336550437` on `main`, which passed.

Next: PR 3 (verify mode), which is the first one that touches deployed behaviour — warn-only by design.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,8 +61,9 @@ These are the constraints most likely to be violated by an agent that skipped `C
 - **Adapters are one service call deep.** Business logic goes in `quantcore/services/`; REST
   routes (`api/routers/*`) and MCP tool bodies (`fastMCPTest/*`) are thin. Service modules never
   import each other or the registry.
-- **Every schema change ships twice** — as a Flyway file under `db/migrations/` *and* mirrored
-  into `init_schema()` in `quantcore/db.py`.
+- **Every schema change touches three files** — a new `db/migrations/V*.sql`, `_SCHEMA` in
+  `quantcore/db.py`, and the regenerated `db/schema_snapshot.json`. This is no longer a
+  convention: `tests/test_schema_parity.py` fails CI if the first two disagree.
 - **BYOK never-log policy:** no API keys, `Authorization` headers, envelopes, decrypted payloads,
   request bodies, or exception dumps containing credentials may reach any log or print. New
   failure paths must add the corresponding log assertion.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -13,7 +13,8 @@ Update the docs when you:
 
 - add, remove, or rename a **service, repository, gateway, or analytics module**
 - add or change a **REST route group**, an **MCP server or tool**, or a **UI page or route**
-- change the **schema** (which also means a Flyway file *and* `init_schema()` — see Migrations)
+- change the **schema** (which means **three** files — a Flyway migration, `_SCHEMA` in
+  `quantcore/db.py`, and `db/schema_snapshot.json`; enforced by CI, see Migrations)
 - change **deployment, CI/CD, environment, or auth** wiring
 - add an **operational script**, or change how an existing one is invoked
 - change a **default, environment variable, or configuration file's role**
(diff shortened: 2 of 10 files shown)


JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** `1ce45bd77ec89211b430a631c0472f7bac50ff2d`
**Verdict:** APPROVE

### Code findings

No actionable bugs, security issues, performance regressions, or missing error handling were found in the changed code. The parity test correctly orders migrations numerically, compares the two schema descriptions, and fails rather than skips when CI cannot reach the database. The migration is idempotent and adds the three tables identified by the parity run.

**Associated issue:** #165 — init_schema() and Flyway both own the deployed schema — flyway info can't be trusted as evidence of prod state

- **Fully met:** The PR implements the issue's selected “keep both, add a drift check” direction by adding a CI-enforced parity test between `init_schema()` and the baseline-plus-migrations schema, and backfills the three tables found missing from the migration set.
- **Partially met:** None.
- **Unmet:** None.
- **Unverifiable:** None.

### Verification

- GitHub Actions run `31338817501`: backend gate passed, frontend gate passed, and secret scan passed.
- Local dependency-free validation: `python -m unittest tests.test_schema_introspect -v` — 22/22 passed.
- Local parity test: ordering and skip-guard tests passed; the database-backed parity case was skipped because no PostgreSQL server was available at `127.0.0.1:5432` in this environment.
- `python -m compileall -q quantcore scripts tests` passed.
- `git diff --check origin/main...HEAD` passed.
- The snapshot check could not run locally for the same unavailable PostgreSQL dependency; the GitHub Actions backend gate passed it on the reviewed head.

PR #174, opened 2026-08-09
feat: QUANTCORE_SCHEMA_MODE — startup checks the schema instead of creating it (#165)

JohnFunkCode:
PR 3 of [`docs/proposals/schema-ownership-plan.md`](docs/proposals/schema-ownership-plan.md), for issue #165. Follows #172 (Evidence) and #173 (Enforcement).

## The problem

On a database Flyway already manages, `init_schema()` running its own `CREATE TABLE` DDL on every application startup makes two systems owners of one schema. This adds the switch that settles it.

## `QUANTCORE_SCHEMA_MODE`

| Mode | Behaviour |
|---|---|
| `create` | Run the 22-table DDL. Today's behaviour, and the escape hatch. |
| `warn` | Introspect, diff against `db/schema_snapshot.json`, log differences, run **no DDL**. |
| `verify` | As `warn`, but raise `SchemaDriftError` on any `MISSING`/`MISMATCH` (`EXTRA` never raises — D4). |
| `auto` *(default)* | `create` where there is no `flyway_schema_history`, otherwise `warn`. |

**This ships warn-only.** Flipping `auto`'s Flyway-managed half to `verify` is PR 4, gated on a clean soak on both projects (D5).

The mode is read at **call** time, not frozen at import like `DB_DSN`, so the escape hatch is one `gcloud run services update --update-env-vars` away. The check emits one greppable line —

```
schema check: mode=auto resolved=warn tables=22 missing=0 mismatch=0 extra=0
```

— plus one line per difference, and no log line carries the DSN.

## Two judgment calls the plan left open

Both resolved toward *don't block startup*:

- **An unrecognized mode** logs an ERROR and falls back to `create`. A typo is most likely made by an operator reaching for the escape hatch mid-incident; failing closed there would deny them the very thing they were reaching for.
- **A missing snapshot file** logs an ERROR in `warn` but raises in `verify` — told to enforce and unable to, fail loudly rather than pass silently.

## Two things the plan got wrong

1. It said `main.py:621` was "the only remaining direct caller outside tests". It was not — `fastMCPTest/options_analysis.py` called `init_schema()` directly too, and would have kept creating tables on a Flyway-managed database. Routed through `ensure_schema()` here as well.
2. `auto` makes the whole suite environment-dependent: the local test database carries a Flyway ledger (`scripts/flyway.sh` defaults to test) and CI's throwaway Postgres does not, so `auto` resolves `warn` locally and `create` in CI. `tests/__init__.py` pins `create`, which also let `test_create_app_does_not_re_run_the_ddl` stay untouched as the plan required.

## Tests

`SchemaModeTest` — 13 cases, database-free. The connection is faked and `describe_schema` stubbed, but **`diff_schemas` is the real one**, so the shipped `MISSING`/`EXTRA`/`MISMATCH` classification is what's under test. Covers both `auto` branches, the `create` escape hatch, a typo'd mode, warn-vs-verify on a missing table, verify leaving the DSN unrecorded so a retry re-checks, `EXTRA` never raising in any mode, the once-per-process gate on the *check* path, a missing snapshot file, and an assertion that no log line carries the DSN.

- Full suite: `Ran 1241 tests … OK (skipped=5)`
- `quantcore/db.py` coverage 97.5%; `diff-cover --fail-under 85` → **100%** on all 52 changed lines

## Live evidence before merge

- Test DB: `auto` → `warn`, `tables=22 missing=0 mismatch=0 extra=0`, DDL skipped.
- Read-only `scripts/schema_check.py --prod`: clean (22 tables, 0 missing, 0 extra), ledger through `V5__watchlist`.

So both projects should log a clean warn line on rollout — that soak is Step 4, and PR 4 is gated on it.

## Docs

`CLAUDE.md` (mode table + rationale) and `readme.md` (env-var bullet, three-file table, the `init_schema()`-races-Flyway paragraph rewritten in past tense). `AGENTS.md` deliberately unchanged — it points at `CLAUDE.md` rather than copying it. Checkpoint row appended to the plan.


JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** `6926dd4d967f457adfaea56f0924ac4618fd9126`
**Verdict:** APPROVE

### Code findings
No blocking or non-blocking code findings. The mode-aware `ensure_schema()` keeps DDL on the create path, performs read-only snapshot checks for Flyway-managed databases, preserves the per-DSN gate, and routes the identified application entry points through `ensure_schema()`.

Issue #165 is associated via the PR title/body reference `#165` (GitHub `closingIssuesReferences` is empty because the issue is already closed).
- **Fully met:** Removes the application startup two-owner behavior for Flyway-managed databases by resolving `auto` to `warn`, while retaining an explicit `create` escape hatch; adds schema drift reporting against the committed snapshot; preserves local/CI bootstrap behavior; updates the documented entry points and tests.
- **Partially met:** None.
- **Unmet:** None.
- **Unverifiable:** None.

### Verification
- GitHub checks: **pass** — Tests + wrapper smoke + OpenAPI surface; **pass** — Frontend tests + typecheck + coverage; **pass** — Secret scan (gitleaks).
- `git diff --check`: passed.
- Python `compileall` on changed Python files: passed.
- Local targeted unittest execution was attempted but could not run because this environment lacks `psycopg2` and `fastmcp`; the corresponding GitHub test check passed.

APPROVE

PR #175, opened 2026-08-10
feat: make schema drift fail the deploy (#165, PR 4 of 4)

### LINKED ISSUE #165: init_schema() and Flyway both own the deployed schema — flyway info can't be trusted as evidence of prod state

JohnFunkCode:
Final PR of [`docs/proposals/schema-ownership-plan.md`](docs/proposals/schema-ownership-plan.md) — the flip. Closes the "two owners of the deployed schema" problem in #165.

## What changes

One line of behaviour: `QUANTCORE_SCHEMA_MODE=auto` now resolves to **`verify`**, not `warn`, on a database carrying a `flyway_schema_history` ledger. On test and prod a `MISSING`/`MISMATCH` now raises `SchemaDriftError` during startup instead of being logged and ignored.

PR #174 soaked this warn-only through a full deploy cycle on both projects — `missing=0 mismatch=0` on each — which was decision D5's gate.

Unchanged where nothing else owns the DDL: local dev, CI, compose and a brand-new instance still get their tables created. `EXTRA` still never raises (D4).

## Why the diff is mostly docs

This changes how a schema change ships:

- **Migrate before you deploy — in both projects.** `deploy.yml` auto-rolls test on every merge to `main`, and test is Flyway-managed too, so an un-migrated change fails the *test* revision before prod ever sees it. (The plan only said "prod"; that was half the exposure.)
- **A migration must now be complete DDL.** A forgotten column used to be invisible because `_SCHEMA` created it at startup anyway. Now it's a `MISSING` line and it fails the deploy.
- **Failing early is the feature.** A revision that fails its startup check never takes traffic — the previous revision keeps serving, instead of drift surfacing hours later as query errors on live traffic.
- **Escape hatch, one command, no code change:** `gcloud run services update … --update-env-vars QUANTCORE_SCHEMA_MODE=create` (`--update-env-vars`, never `--set-env-vars`).

That's stated in `CLAUDE.md`, `readme.md` and `AGENTS.md`, in `scripts/flyway.sh`'s header (which loses its "shape converges on startup" claim), and in the `_resolve_schema_mode()` / `ensure_schema()` docstrings — where an operator reading the code mid-incident will actually find it.

## One deviation from the plan

The migrate-first step went into [`docs/operations/prod-promotion.md`](docs/operations/prod-promotion.md) (now step 2 of "Promote (the actual procedure)", with `scripts/schema_check.py --prod` as confirmation), **not** `prod-rollout-plan.md` as the plan's documentation table said — that file is a closed checkpoint log for the one-time buildout, not the runbook operators follow. Both the plan's table and its Step 2 are corrected to match.

## Verification

- Full suite: `Ran 1242 tests … OK (skipped=5)`
- `diff-cover coverage.xml --compare-branch=main --fail-under 85` → **100%**, 0 missing lines
- `quantcore/db.py` at 97.5% statement coverage
- New test `test_auto_enforces_on_a_flyway_managed_database` pins the flip; `test_extras_never_raise_in_any_mode` now covers `auto` alongside `warn`/`verify`

## Follow-up (not in this PR)

Plan Step 3 — comment on and close #165, plus the follow-up comment `legacy-report-retirement-plan.md` Part H1 asks for. Held until this merges.


JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** b5af72d2b24331805317821b35ba794aab5adbf5
**Verdict:** APPROVE

### Code findings
No Critical, Warning, or Suggestion findings. The `auto` flip is narrowly implemented, preserves `create` for unmanaged databases, retains the documented EXTRA-object policy, and adds regression coverage for the enforcement path.

### Issue objective assessment — #165
- **Fully met:** Remove the application as a competing schema owner on Flyway-managed databases; make schema drift visible and fail startup rather than silently recreating objects; preserve bootstrap behavior where no Flyway ledger exists; make migrate-before-deploy ordering explicit in the operator documentation.
- **Partially met:** None.
- **Unmet:** None.
- **Unverifiable:** None.

### Verification
- GitHub PR checks: Tests + wrapper smoke + OpenAPI surface **passed**; Frontend tests + typecheck + coverage **passed**; Secret scan **passed**. Deploy and credential-check jobs were skipped by workflow conditions.
- Local targeted `tests.test_schema_bootstrap`: 19 passed; the DB-dependent `test_create_app_does_not_re_run_the_ddl` could not run because no PostgreSQL server was available at localhost:5432.
- Local full unittest discovery: 1129 tests, 6 skipped, 142 errors, all attributable to the unavailable localhost:5432 PostgreSQL service (plus import failures cascading from app bootstrap); not treated as PR findings.
- Python compilation: passed.

*Reviewed by Hermes Agent*

PR #177, opened 2026-08-10
refactor: extract the nav/route seam out of App.tsx (#147 PR 0)

JohnFunkCode:
PR 0 of nine for #147, per the Sequencing section of
[`docs/proposals/legacy-report-retirement-plan.md`](docs/proposals/legacy-report-retirement-plan.md).
This is the seam PR — mitigation 1. It ships no feature; it makes the eight PRs behind it
stop colliding.

## Why

`App.tsx` carried two parallel lists — a `navItems` array for the nav buttons and a
`<Routes>` block for the router — so adding a page meant two edits in two places, with
nothing stopping them from disagreeing. #147 adds four pages across four PRs on two
parallel tracks, which would have made `App.tsx` the hottest file in the series and every
one of those PRs a merge-conflict candidate against the same two lists.

## What changed

- **`frontend/src/navigation.tsx` (new)** — one exported `routes` array of
  `{ path, label, icon, element, group?, nav? }` that both render sites map over, so adding
  a page is a one-object append. `navItems` and `isPathActive()` derive from it.
  The array is typed so the nav guarantee is structural rather than a convention:
  `NavRoute` requires an `icon`, `DetailRoute` requires `nav: false` and forbids one, and
  `navItems` narrows through a type predicate. `group` is declared but unused — Part G2
  collapses the flat button row into dropdowns and derives them from it, which should be a
  render-site change rather than another list rewrite.
- **`frontend/src/App.tsx`** — 199 → 167 lines; 16 imports dropped. Keeps the wordmark, the
  per-button neon treatment, the Sidekick and theme toggles, the hide-not-unmount fullscreen
  container, and the `not_provisioned` early return (#126).
- **`frontend/src/navigation.test.tsx` (new)** — 12 tests: no duplicate paths, `/` stays the
  index, the hidden set is exactly the two drill-down routes, nav-bar contents and ordering,
  no parameterized path in the bar, and the `'/'` guard in `isPathActive` (without it,
  `startsWith` lights up Portfolio on every page).
- **`tests/test_schema_parity.py`** — `test_migration_versions_are_unique`. Two branches
  both numbering their migration V7 is the collision git allows silently: the merge yields
  two V7 files rather than a conflict, and Flyway only refuses at rollout time, against a
  deployed database. It lands in `MigrationOrderTests` because that class is pure and
  already runs in the `gate` job without a database.
- **`docs/proposals/legacy-report-retirement-plan.md`** — the pending Part H1 corrections:
[...]

Issue #147, opened 2026-07-27
Decouple the unused HTML report and S3 upload from main.py, and surface the nightly report in QuantUI


JohnFunkCode:
Raised during the design discussion for #126 (per-user external portfolios).

## Problem

`main.py` (658 lines, run daily as the `quantcore-report` Cloud Run Job) mixes
two unrelated responsibilities, and only one of them is currently consumed.

**Consumed — the job's real value:**

- OHLCV cache population (price/metrics fetch for portfolio + watchlist symbols)
- Options-chain snapshot capture (`main.py:651`), which is what makes the
  open-interest time series and daily GEX regime history accumulate
- Discord notifications via `notifier.py` — moving-average violations, loss
  alerts, harvester rung hits, sentiment flips, options position alerts

**Not currently accessible to users — the HTML report path:**

- `create_portfolio_charts()` (`main.py:27`) — matplotlib figures
- `fig_to_base64()` (`main.py:19`) — base64 embedding
- `create_portfolio_html()` / `create_template_file()` (`main.py:130`, `:227`)
  — Jinja2 rendering against `templates/portfolio_template.html`
- `save_html_to_s3()` (`main.py:553`) — boto3 upload to `BUCKET_NAME`/`BUCKET_KEY`

The generated report is **not accessible by users today** because the html is that would be stored in an S3 bucket is not getting stored because the bucket secrets are not provided anywhere.  However we want to surface this very useful information to all users. It predates the QuantUI
front end, which now covers everything the report showed — except that it is
still generated, still rendered, and still uploaded on every nightly run.

## Consequences

- The daily job carries matplotlib, Jinja2, and boto3 purely for a dead artifact,
  inflating `Dockerfile.report` build time and image size.
- A failure in the report path can fail the job and take the *useful* side
  effects (cache population, options snapshots, notifications) down with it,
  even though nothing depends on the report.
- The valuable parts are buried inside a module named after the dead part, which
  makes the job's actual purpose non-obvious.

## Scope

**1. Decouple.** Separate the nightly job into its consumed responsibilities
(data capture + notifications) and the report-rendering path, so a report
failure cannot break data capture. Per `docs/proposals/architectural-standard-v2.md`,
the logic belongs in the services layer with `main.py` as a thin entry point;
today `main.py` holds substantial presentation logic directly.

**2. Surface the report in the UI** — the outcome we actually want. Rather than
retiring the report, give it a home in QuantUI so the nightly snapshot is
[...]

PR #178, opened 2026-08-10
feat: nightly fundamentals warmer + extract the HTML report from main.py (#147)

JohnFunkCode:
PR **1 of 9** in [`docs/proposals/legacy-report-retirement-plan.md`](docs/proposals/legacy-report-retirement-plan.md), covering **Parts E and F** of issue #147. Follows PR 0 (#177).

## Part E — nightly fundamentals warming

`main.py`'s `__main__` gains a final step that refreshes the fundamentals cache over the same universe the options capture uses (John's positions + the global watchlist + every other owner's positions), **oldest-`fetched_at` first**, under a wall-clock budget.

A cold pass is ~10 serial yfinance calls per symbol and is *expected* not to finish — the universe converges over a few nights, which is what makes the ordering load-bearing rather than cosmetic. Ordering comes from the new `FundamentalsService.cache_freshness()`.

**The isolation is the point.** The step runs last, after the notifications and options capture have already succeeded, wrapped in an outer `try/except` that never raises — a warmer failure must not retroactively fail a run whose alerts already went out. That silence is then made loud: `alert_if_fundamentals_stale()` re-reads freshness **after** warming and fires a Discord alarm on either of two independent triggers.

The two triggers catch different failures: **coverage** catches a warmer that is not keeping up at all; the **age ceiling** catches a few symbols that fail every single night while the average stays healthy.

The embed title carries the date, because `send_notifications()` dedupes on title against `notification.log` — a static title would alarm once and then go quiet while the cache stayed stale.

| Env var | Default | Meaning |
|---|---|---|
| `FUNDAMENTALS_WARM_BUDGET_SECONDS` | `900` | wall-clock budget for the warming pass |
| `FUNDAMENTALS_STALE_COVERAGE_FLOOR` | `0.80` | alarm below this in-TTL fraction |
| `FUNDAMENTALS_STALE_MAX_AGE_HOURS` | `168` | alarm above this oldest age |

An unparseable value logs a warning and falls back to the default — a typo in a Cloud Run env var must not silently disarm the alarm.

## Part F — the legacy HTML report leaves `main.py`

The four rendering functions move **verbatim** into `scripts/generate_portfolio_report.py` (`main.py` 704 → 282 lines), which the Pi now runs via `runOnPi.sh`. The Pi runs the script and **not** `main.py`, because the Cloud Run Job already sends the notifications and captures the snapshots — running both would double every alert.

Two things could not move verbatim, and both are covered by tests:

- **The Jinja loader path** was `Path(__file__).parent`, which meant the repo root while the function lived in `main.py` and would mean `scripts/` here — silently generating a second, divergent copy of the template. Now `REPO_ROOT / "templates"`.
- **`--publish` now fails loudly** when `BUCKET_NAME`/`BUCKET_KEY` are missing, checked up front before minutes of price fetching. The old code returned `None` and let the caller report success, which is how the public page could stop updating unnoticed — the original #147 defect.

That split lets **matplotlib, jinja2, and boto3 leave `requirements-base.txt`**, so no container image carries them any more. `requirements-dev.txt` pulls `requirements-report.txt` back in because CI still needs it — the suite now imports the report script, and untested is exactly how that code drifted into being un-runnable in the first place. Being the CI-only file, it puts nothing back into any image.

## Verification

- Full backend suite: **1281 tests, `OK (skipped=5)`**
- The four PR-1 test modules together: 43 tests, `OK`
- All **nine container entrypoints** (`main`, `api.main`, the 7 MCP servers) still import with `matplotlib`/`jinja2`/`boto3` blocked at the meta-path — the lean images genuinely survive the dependency removal

## Notes for the reviewer

- **#147 is commented on but deliberately not closed**, against Part F's instruction: it is the tracker for all nine PRs and now carries the mitigation-5 reservations. Closing it here would close the tracker two PRs into the plan.
- **`docs/openapi-surface.txt` is not regenerated** — this PR adds no REST route.
- Part F's docs bullet also asks for `/watchlist`, `/fundamentals`, `WatchlistService`, `scope`, and `quantcore/analytics/returns.py`; none of those exist until PRs 3–5, so the docs were scoped to what is true today.
- **Found in passing, out of scope here:** `FundamentalsRepository.stats()` returns the password-bearing `QUANTCORE_DB_DSN` on both its success and error paths, and `FundamentalsService.get_cache_stats()` passes it straight through to the `get_cache_stats` MCP tool. Being fixed separately.


JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** `b832648b9b381f1b95d48b7d64cd5c185af88690`
**Verdict:** REQUEST_CHANGES

### Code findings

- **main.py:107** — **Warning** — `warm_fundamentals_cache()` uses `DEFAULT_WARM_BUDGET_SECONDS` whenever `budget_seconds` is omitted, but `run_fundamentals_warming()` omits it and never reads `FUNDAMENTALS_WARM_BUDGET_SECONDS`, so the documented environment override is ignored and the job can run for the hard-coded 900 seconds despite deployment configuration.
  **Fix:** Resolve the default through `_env_float(WARM_BUDGET_SECONDS_ENV, DEFAULT_WARM_BUDGET_SECONDS)` (or pass that resolved value from `run_fundamentals_warming()`), and add a test proving the environment variable changes the effective budget.

- **tests/test_generate_portfolio_report.py:112** — **Suggestion** — The new test asserts that `report.REPO_ROOT.name` is exactly `StockPortfolioManager`, making the suite depend on the checkout directory name; this failed in a valid checkout at `/tmp/spm-pr178`.
  **Fix:** Assert the path relationship (`REPO_ROOT / "templates"` and the expected repository files) instead of a directory basename, or remove this environment-specific assertion.

Issue #147 is explicitly referenced by the PR and is the associated tracker issue.

- **Partially met:** Decoupling the daily job from HTML rendering/S3 and isolating the long-running fundamentals warmer are implemented, but the issue's broader architectural objective of moving remaining orchestration into the services layer is explicitly deferred by the issue/PR notes.
- **Unmet:** Surface the nightly report in QuantUI; this PR extracts a manually runnable legacy report script but does not add the UI page or REST/service path.
- **Fully met:** Preserve a manually runnable report/publish path outside `main.py`, with explicit S3 credential failure and report dependencies removed from the lean container dependency set.

### Verification

- Target tests: `python -m unittest tests.test_fundamentals_cache_freshness tests.test_fundamentals_warmer tests.test_generate_portfolio_report` — **37 passed, 1 failed**; the failure is the checkout-directory assertion noted above.
- Full local discovery: `1012 tests`, **1 failure and 23 errors**. The failure is the same checkout-directory assertion; the errors are environment-blocked database tests (no test DSN/PostgreSQL at localhost), not attributed to this PR. Five tests were skipped.
- Manual verification: with `FUNDAMENTALS_WARM_BUDGET_SECONDS=1`, the warmer still reported and used `budget 900s`, confirming the configuration bug.

Please fix the budget override and make the new path test independent of checkout directory before re-review.

JohnFunkCode:
Thanks — both code findings were real and are fixed in `90bbb4a`.

## 1. `FUNDAMENTALS_WARM_BUDGET_SECONDS` was dead (main.py:107)

Correct, and worse than a missed override: the knob was declared, defaulted, and written into three documents while never being read. `warm_fundamentals_cache()` resolved its default as `DEFAULT_WARM_BUDGET_SECONDS if budget_seconds is None else …`, never through `_env_float`.

What kept it hidden is that the two *alarm* thresholds ten lines below **do** read the environment, so the block looked internally consistent on inspection. And since the one production call site (`run_fundamentals_warming`) passes no budget, setting the variable on the Cloud Run Job would have changed nothing — exactly what your `FUNDAMENTALS_WARM_BUDGET_SECONDS=1` run showed.

Resolution moved **inside** the function rather than to the call site, because a default resolved at the call site is only as good as the call sites, and the one that matters here passes nothing. Three tests pin the directions:

- `test_the_budget_comes_from_the_environment_when_not_passed` — the env var changes the effective budget
- `test_an_explicit_budget_still_beats_the_environment` — an explicit argument still wins
- `test_the_env_budget_reaches_the_warmer_through_this_caller` — the value arrives via the real caller, which is the actual regression

Your manual repro now reports `budget 1s`.

## 2. Checkout-directory assertion (tests/test_generate_portfolio_report.py:112)

Also right. Replaced with the path relationship: `REPO_ROOT` is the script's parent and contains `main.py` and `templates/`, which is what the test was actually about.

Reproduced rather than taken on faith — in a worktree at `spm-pr178-checkname`, `REPO_ROOT.name` is `spm-pr178-checkname`, the old assertion evaluates `False`, and the 9 tests pass under the new one. Worth noting for anyone reading later: **CI passing was not evidence here**, because Actions checks out into a directory named for the repo — the most conventional layout there is.

## On the objective assessment

One clarification rather than a change: "**Unmet:** Surface the nightly report in QuantUI" is accurate against issue #147 as a whole, but #147 is the tracker for a **nine-PR plan** ([`docs/proposals/legacy-report-retirement-plan.md`](docs/proposals/legacy-report-retirement-plan.md)), and this is PR 1 of 9, covering Parts E and F only. The QuantUI page, its REST route, and the service path are Parts A–D and G, landing in PRs 2–8. Likewise the "remaining orchestration into the services layer" objective is deferred deliberately and recorded in the plan, not overlooked.

That is also why **#147 stays open** rather than being closed by this PR — closing the tracker two PRs into a nine-PR plan would lose the remaining scope.

## Verification on `90bbb4a`

- Full backend suite: **1284 tests, `OK (skipped=5)`** (up 3 — the new budget tests, no other movement)
- Target modules: `tests.test_fundamentals_cache_freshness tests.test_fundamentals_warmer tests.test_generate_portfolio_report` — all green, including from a differently-named checkout

Your run of 1012 tests with 23 errors was the missing test PostgreSQL — `tests/__init__.py` swaps in the test DSN, so the DB-backed suites need a reachable instance. Ready for re-review.

PR #182, opened 2026-08-10
feat: stacked allocation bar on the Portfolio page (#147 Part A)

JohnFunkCode:
PR 2 of 9 for issue #147 (legacy HTML report retirement). Part A in the plan's Sequencing — the first of the report's four visuals to move into QuantUI.

## What it does

One bar per symbol on the Portfolio page: an orange cost-basis block from zero, with the gain stacked **above** it in green, or the loss cut **down** into it in red.

That last detail is the legacy matplotlib semantics reproduced exactly (`scripts/generate_portfolio_report.py:95-115`) — the red block overlays the *top* of the orange one, so the visible top of every bar is the position's current value whether it is up or down. Bars sort biggest-first, since the chart's job is relative position size. A test pins the overlay geometry so a future refactor can't quietly turn a loser into a shorter bar.

## Where the math lives

Server-side, per Rule 8.4. New `quantcore.analytics.portfolio_math.allocation_segments()` returns `bar_base`/`bar_gain`/`bar_loss`; `PortfolioService.symbol_rows()` spreads them into each row; `SymbolRowResponse` carries them. The chart does pixel scaling and nothing else.

`bar_loss` is deliberately negative, and all three are `null` for a symbol with no priced lot — which the chart **drops** rather than drawing as a zero-height bar, because a flat bar reads as a worthless position rather than an unknown one.

## GenUI

Ships as `portfolio_allocation` with an empty prop spec in **both** registries (`quantcore/services/chat_tools.py` + `frontend/src/chat/componentRegistry.tsx`), plus the `show_component` tool prose. It inherits `portfolio_table`'s no-props rule for the same reason — decision #20: a model-settable `owner` prop would be a cross-user read. Parity cases mirrored on both sides.

The chat card self-fetches on the shared `['portfolio-symbols']` query key, so it costs no extra request; the Portfolio page passes down the rows it already has.

Both mounts guard the nothing-to-draw case rather than leaving an empty panel: the page skips the `Paper` entirely, the card says "No priced positions to chart yet".

## Test note

Six existing `PortfolioPage` tests needed scoping to the holdings table — the chart's x axis now labels the same tickers, so a page-wide `getByText('INTC')` became ambiguous. Two page tests added: the chart mounts from the already-fetched rows (asserting exactly **one** `/api/portfolio/symbols` call), and stays off the page when nothing is priced.

## Verification

- Frontend **468 tests green** (up 18), `tsc --noEmit` clean
- Coverage **89.95 / 87.96 / 87.82 / 71.26** vs floors 89 / 87 / 86 / 70 — floors **not** raised; that happens once in Part G (plan mitigation 2)
- Backend `portfolio_math` / `portfolio_service` / `chat_tools` / `chat_protocol` — **122 tests OK**
- `scripts/check_openapi_snapshot.py` matches (106 routes)

Also folds in housekeeping: the plan's PR 1 checkpoint row now records the merge (#178, `5a21bb5`).


JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** `bf58207af7e930f48ccd9f822e9db076171f5744`
**Verdict:** APPROVE

### Code findings

No actionable bugs, security issues, performance regressions, or test deficiencies found in the reviewed patch.

Associated issue: #147 — Decouple the unused HTML report and S3 upload from main.py, and surface the nightly report in QuantUI.

- **Fully met:** Scoped Part A objective — the portfolio allocation visualization is implemented with server-side segment math, rendered on the Portfolio page, registered in both GenUI registries with no cross-user props, and covered by backend/frontend tests.
- **Partially met:** None for the scoped Part A deliverable.
- **Unmet:** None for the scoped Part A deliverable. The broader issue remains open by design and is addressed by the series of follow-up PRs.
- **Unverifiable:** None for the scoped Part A deliverable.

### Verification

- GitHub checks: Frontend tests + typecheck + coverage — **pass**; Tests + wrapper smoke + OpenAPI surface — **pass**; Secret scan — **pass**. Build/deploy and deploy-credentials checks were skipped by CI.
- Local `git diff --check` and Python `compileall` — **pass**.
- Local targeted Python unit tests could not run because the checkout environment lacks `pytest` and the repository requires `QUANTCORE_DB_DSN`; this is recorded as an unavailable local check, not a code failure.
- Existing GitHub reviews/comments were checked; no prior Hermes review for this head SHA was present.

PR #183, opened 2026-08-10
feat: watchlist returns + fundamentals in six queries (#147 Parts B1–B3, B5)

JohnFunkCode:
PR 3 of nine for [#147](https://github.com/JohnFunkCode/StockPortfolioManager/issues/147), following the Sequencing in [`docs/proposals/legacy-report-retirement-plan.md`](docs/proposals/legacy-report-retirement-plan.md). Parts **B1, B2, B3, B5** — B4 is deliberately held back for PR 5, where it ships with Part D.

This replaces the nightly `generate_watchlist_fundamentals_report.py` HTML with a REST route that reads cache only, which is what lets the same table become a page instead of a batch job. **No UI yet** — the `/watchlist` page is PR 4.

## B1 — `quantcore/analytics/returns.py`

Not new math: **one correct copy replacing three inconsistent ones**. Everything is close-to-close, which is the report script's convention and the correct one. The module documents the three defects in `portfolio/metrics.py` that made consolidation the only option rather than reuse in place:

- it computes **open-to-close** (`Open.iloc[-6]` → `Close.iloc[-1]`), so a 5-day return silently includes the sixth day's overnight gap;
- it anchors YTD by counting business days with `bdate_range`, which counts market holidays as business days — so the positional index drifts off the actual first trading bar of the year;
- it divides by `iloc[0]` of a **two-year** frame and labels the result `one_year_return`.

`portfolio/metrics.py` is left alone, with a comment saying why: it feeds the preserved legacy report script, and changing it here would silently move numbers on a page nobody is watching.

`normalize_market_cap` returns **three** fields — the native figure, its currency, and USD only when a rate is known — rather than one converted number. It does **not** fetch FX: the only FX source in the repo (`portfolio/money.py`) is a network call to open.er-api.com, which would destroy the zero-network property below. The bug it exists for is SK hynix's ₩1,456T rendering as `1456.62T` in a column of dollars and sorting to the top.

## B2 — `PricesService.bulk_period_returns`

Modelled on `screen_securities` (one `daily_bars_for_symbols` read, group in memory, then pure analytics), deliberately **not** on `PortfolioService._period_returns`, which loops per symbol. Two details worth a reviewer's eye:

- the `ts` epoch → `date` conversion happens in the service, so `analytics.returns` never sees a repository row shape (Rule 8.4);
- the calendar anchor defaults to the **newest bar, not today** — on a Sunday, or with a symbol whose cache is a few days behind, "today" would silently widen the YTD/1y windows past the data.

## B3 — service composition

`FundamentalsService.get_all_latest` / `cache_ttl_seconds` are thin passthroughs placed on the *service* rather than letting `WatchlistService` reach into `FundamentalsRepository`: crossing into another domain's repository would mean re-deriving that domain's staleness rule at each call site. `ttl_seconds()` is an env read, not a query — which is what makes the count below exact.

`WatchlistService.returns_and_fundamentals` composes those services (the `RecommendationsService` precedent, arch-v2 §5.2), both injected **optionally** so the CRUD half stays constructible from a bare repository (import scripts, existing tests).

**Six queries, zero network:** one `list_entries`, one `bulk_period_returns`, four `get_all_latest`. The count is constant in the size of the watchlist.

Cached fundamentals are shown **stale rather than dropped**, with `fundamentals_stale` / `fundamentals_age_hours` on the row: a table that loses rows overnight reads as "we stopped watching that" instead of "this number is a week old". A symbol never scored simply has null fundamentals — never zeros, since a zero score and an unscored symbol sort to very different places.

## B5 — `GET /api/watchlist/fundamentals`

Declared **before** `/watchlist/{ticker}` — FastAPI matches in declaration order, so the reverse would let the DELETE path template swallow `fundamentals` as a ticker. Nothing enforces that but the comment above the route. No `?owner=`; the watchlist is global ([#83](https://github.com/JohnFunkCode/StockPortfolioManager/issues/83)).

Unlike PR 2, this one **does** move `docs/openapi-surface.txt` (106 → 107): the gate is route-level, so new response fields are invisible to it but a new route is not.

## Verification

| Check | Result |
|---|---|
| Full backend suite | 1334 tests, 5 skipped. One error — `test_schema_parity`'s `DROP DATABASE … WITH (FORCE)` **teardown**, from two suites racing on one test database; green in isolation (8 tests OK) |
| `tests/test_returns_analytics.py` (new) | 23 passed |
| `tests/test_watchlist_service.py` | 21 passed |
| OpenAPI surface gate | passes, 107 routes |
| Schema three-files rule | N/A — no schema change |
| vitest floors | untouched — no frontend here (plan mitigation 2 raises them once, in Part G) |

The six-queries test is built against the **real** `PricesService` and `FundamentalsService`, with counting fakes beneath them and an `ExplodingYFinance` gateway whose `__getattr__` raises on any attribute — because the easy wrong version of this method is a per-symbol loop that still returns the right answer. It asserts both halves at 1, 3 and 25 symbols to pin that the count is constant in list size.

Smoke-tested against a live database: **228 rows, `stale_count: 0`, `unscored_count: 0`**, first row `SBMO.AS` (EUR) carrying `market_cap: 5344592896.0`, `market_cap_currency: "EUR"`, `market_cap_usd: null` — the currency bug behaving exactly as designed rather than inventing a dollar figure. `DELETE /api/watchlist/ZZZZ` still returns 404, confirming the new GET does not shadow the DELETE path template.

## Docs

`CLAUDE.md` (analytics module list, the `WatchlistService` description, the watchlist route list and its ordering constraint), `readme.md` (a human-facing paragraph on the endpoint), and the plan's checkpoint log — which also corrects PR 2's row from "In review" to Merged.


JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** `4ef4820f9b971a9038662b33a9689376f4596262`
**Verdict:** REQUEST_CHANGES

- **Fully met:** The scoped Part B1/B2/B3/B5 deliverables described in the PR are implemented: pure return analytics, bulk cache-only returns, service composition, and the REST route.
- **Partially met:** Issue #147's objective to surface the nightly-report information in QuantUI: this PR adds the REST data endpoint, but no UI page is included yet.
- **Unmet:** Issue #147's objective to decouple the nightly job's report-rendering/upload path from the consumed data-capture and notification responsibilities; this PR does not change that job path.
- **Unverifiable:** The issue's chart-fate decision and any future historical-report/archive behavior are not addressed by this scoped PR.

### Code findings
- **quantcore/services/prices.py:258** — **Severity — Warning** — When `as_of` is supplied, `closes` and `dated` still include bars newer than that date, so `price`, `price_as_of`, `bars`, and every trailing return are calculated from future data while only YTD/1y honor the requested anchor. **Fix** — filter each symbol's dated bars to `d <= as_of` before constructing `closes` and deriving the row metadata; retain the newest retained bar as the effective as-of date (and add a regression test with bars after the requested date).

### Verification
- `unittest` return-analytics tests: **23 passed**.
- `compileall`: **passed**.
- `git diff --check`: **passed**.
- `pytest tests/test_returns_analytics.py tests/test_watchlist_service.py -q`: unavailable because `pytest` is not installed in the review environment.
- The watchlist-service unittest and OpenAPI snapshot check were unavailable because the review environment lacks `psycopg2`.
- GitHub checks at review time: secret scan passed; backend/wrapper/OpenAPI and frontend/typecheck/coverage checks were still pending.

PR #184, opened 2026-08-10
fix: resolve a watchlist entry's currency from the exchange, not the caller

JohnFunkCode:
## The defect

`watchlist.yaml` declared **ASSA-B.ST** (Stockholm), **AUTO.OL** (Oslo) and **NIB.F** (Frankfurt) as `USD`. That isn't a missing number on the watchlist fundamentals page — it's a *wrong* one: a SEK market cap rendered with a dollar sign. Three wrong out of a handful of foreign names is enough to conclude that asking a human to type a currency next to a ticker doesn't work.

So this takes it out of their hands rather than validating what they typed.

## What changed

`WatchlistService.add_entry` now resolves the currency itself, via `YFinanceGateway.ticker_info` → `info["currency"]`. That is the **trading** currency, which is the unit yfinance's `marketCap` is quoted in (price × shares) — `financialCurrency` is the reporting currency and can differ, so using it mislabels the cap just as surely.

The `currency` argument stays in the signature but is demoted to a **fallback**, used only when the lookup comes back empty. A disagreement between the two is logged.

**The lookup fails soft, never closed.** Yahoo being slow or down is not a reason to refuse to watch a symbol — a miss falls back to the supplied value and warns; it never raises. The warning is gated on a gateway actually being wired, so import scripts and unit tests constructing a bare service stay silent (that's the design there, not a degraded path).

### Consequences, each with a test guarding it

| Change | Guard |
|---|---|
| `add_to_watchlist` on the MCP wrapper has **no `currency` parameter** at all | `tests/test_mcp_seam.py::test_add_to_watchlist_takes_no_currency_argument` |
| The Add Security dialog shows its currency picker on the **Portfolio tab only** | `AddSecurityDialog.test.tsx` (both directions) |
| `POST /api/watchlist` returns `{symbol, destination, currency}` — what was *stored* | `tests/test_api_smoke.py` |

Positions genuinely need a caller-supplied currency (the portfolio service does no lookup), so the picker was scoped rather than deleted.

`_compute_fundamental_score`'s payload also now carries `market_cap_currency`, so a consumer can label the cap correctly.

### The rows already in the table

New-add correctness doesn't touch the ~228 entries seeded from YAML, so **`scripts/repair_watchlist_currency.py`** drives a new `WatchlistService.resync_currencies`:

```bash
python scripts/repair_watchlist_currency.py            # dry run: prints the diff
python scripts/repair_watchlist_currency.py --apply
```

Dry run by default — the input is a live network lookup, and a bad afternoon at Yahoo shouldn't be able to rewrite the whole table unattended. One request per symbol (a full pass takes a few minutes), `--symbols A,B` to scope, and it refuses the production DB in `.env` without `--allow-prod`. **A symbol whose lookup fails is reported and left alone** — a failed lookup is not evidence the stored value is wrong.

## Verification

- `tests.test_watchlist_service` — 26 OK (12 new)
- `tests.test_mcp_seam` — 28 OK
- `tests.test_api_smoke` + fundamentals + portfolio-manager suites — OK
- `vitest run` — **71 files, 470 tests passed**
- OpenAPI surface regenerated: 106 routes, **no diff** (the snapshot tracks paths and methods, not response models)

## Two things to know before merging

1. **The consume half isn't in this PR.** `market_cap_currency` is now persisted, but `WatchlistService.returns_and_fundamentals` — which would read it — only exists on #183. That one-liner has to land there or later.
2. **`market_cap_usd` stays `None` for non-USD names** until there's an FX source. Correct-but-blank beats wrong-and-confident; a separate `fx_rates` issue would close it.

Docs updated in the same PR per `CLAUDE.md`: the watchlist section in `CLAUDE.md`, the `watchlist.yaml` config bullet, and a "Currency is looked up, not asked for" subsection in `readme.md`.


JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** `67a401eaf21e69ac0b83a2697a02df59cf9ca428`
**Verdict:** REQUEST_CHANGES

- **Fully met:** No associated issue found. The PR's stated behavior is covered by the implementation and added tests, subject to the warning below.
- **Partially met:** None.
- **Unmet:** None.
- **Unverifiable:** None.

### Code findings
- **quantcore/services/watchlist.py:136** — **Severity — Warning** — `add_entry` now calls `YFinanceGateway.ticker_info` on the synchronous watchlist-add path, but that gateway's `ThreadPoolExecutor` context manager waits for its worker during cleanup after `future.result(timeout=...)` times out, so a hung Yahoo request can still hold the HTTP request indefinitely instead of failing soft as documented. The new call makes this existing gateway timeout flaw user-visible on every add.
  **Fix** — Make the gateway timeout genuinely bounded (for example, avoid a context manager that performs blocking executor cleanup on the timeout path and arrange cancellation/worker isolation), or use a non-blocking/asynchronous lookup with a bounded fallback before returning from the add request; add a regression test proving the add returns within the timeout when the info call hangs.

### Verification
- `python -m compileall -q api fastMCPTest quantcore scripts tests`: passed.
- `git diff --check origin/main...HEAD`: passed.
- `python -m unittest tests.test_watchlist_service tests.test_mcp_seam`: unavailable; environment lacks `psycopg2` and `fastmcp`.
- Frontend dialog test: unavailable; `frontend/node_modules` is not present.
- GitHub checks: no checks reported for this head at review time.

JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** `14eae734e51fe444af789b5647c1e7cdae82bd04`
**Verdict:** REQUEST_CHANGES

- **Fully met:** No associated issue found. The PR references `#183`, which is a pull request rather than an issue, and does not use a closing/reference form that associates an issue.
- **Partially met:** None.
- **Unmet:** None.
- **Unverifiable:** None.

### Code findings
- **quantcore/services/watchlist.py:302** — **Severity — Warning** — `add_entry` now puts `YFinanceGateway.ticker_info` directly on the synchronous watchlist-add path (`looked_up = self.resolve_currency(row["symbol"])`), but `ticker_info` uses a `ThreadPoolExecutor` context manager whose `__exit__` waits for the worker after `future.result(timeout=...)` expires; a hung Yahoo request can therefore keep the HTTP request blocked instead of failing soft as documented.
  **Fix** — Make the timeout genuinely bounded (for example, avoid blocking executor cleanup on the timeout path and isolate/abandon the worker safely, or use an asynchronous/non-blocking lookup with a bounded fallback), and add a regression test proving an add returns promptly when the info call hangs.

- **quantcore/repositories/watchlist_repository.py:189-200** — **Severity — Suggestion** — The required CI diff-coverage check fails at 82% (below the repository's 85% threshold), with the new `set_currency` database write path's success/rollback lines uncovered.
  **Fix** — Add repository-level tests for `set_currency` covering the updated-row and zero-row cases, plus the execute/commit failure rollback path, so the required coverage gate passes.

### Verification
- `python -m compileall -q api fastMCPTest quantcore scripts tests`: passed.
- `git diff --check origin/main...HEAD`: passed.
- GitHub checks: frontend tests/typecheck/coverage passed; secret scan passed; backend/wrapper/OpenAPI check failed because diff coverage was 82% versus the required 85% (missing `set_currency` lines 189-191 and 195-200). Deployment jobs were skipped because deploy credentials were not configured.
- `python -m unittest tests.test_watchlist_service tests.test_mcp_seam`: unavailable in this environment because `psycopg2` and `fastmcp` are not installed.
- Frontend tests could not be run locally because `frontend/node_modules` is absent.

Issue #185, opened 2026-08-10
Add an FX rate source so market_cap_usd is populated for non-USD securities


JohnFunkCode:
`market_cap_usd` is `null` for every non-USD security because the repo has no FX source. The
field, the seam that would fill it, and the tests around it all exist already — only the rate
supply is missing.

### What it costs today

Against the test database's 228-row watchlist:

| | rows |
|---|---|
| USD | 210 |
| EUR 5, KRW 4, HKD 2, CNY 2, SEK/SGD/NOK/JPY/DKK 1 each | 18 |

18 rows carry a native market cap and no USD equivalent, across 9 currencies. This is not
cosmetic: KRW is off by ~1400x and JPY by ~150x against a dollar column, which is the SK hynix
`1456.62T` bug that `normalize_market_cap`'s three-field split (`market_cap`,
`market_cap_currency`, `market_cap_usd`) was written to stop. The current behaviour is correct
but incomplete — a missing number that admits it, rather than a wrong one.

`quantcore/analytics/returns.py:130` already accepts `fx_rate` (USD per one unit of `currency`)
and nothing passes it. `WatchlistService.returns_and_fundamentals` is the only caller.

### Proposed shape

- **`fx_rates` table** — `as_of date, quote_currency text, usd_per_unit numeric, source text,
  fetched_at timestamptz`, PK `(as_of, quote_currency)`. Schema change means the three-files rule
  applies: a new `db/migrations/V*.sql`, `_SCHEMA` in `quantcore/db.py`, and
  `db/schema_snapshot.json`.
- **`FxGateway`** over `open.er-api.com` (already in the allowlist). One call returns every rate
  against USD; invert `rates[X]` to get USD-per-unit. Refresh as a step in `main.py` **after**
  notifications, preserving the daily job's deliberate isolation ordering — one HTTP call a night,
  and a warmer-style outer `try/except` so an FX failure cannot retroactively fail a run whose
  alerts already went out.
- **`FxService.usd_rates(currencies)`** reads the table, never the network. `returns_and_fundamentals`
  goes 6 -> **7** queries, still constant in list size and still zero-network on the read path;
  `tests/test_watchlist_service.py`'s query-count guard moves from 6 to 7.
- **Staleness, bilge-pump style.** Accept the newest rate up to N days old; past that return `None`
  rather than a stale figure, so the failure stays visible as a blank instead of quietly drifting.
[...]

PR #186, opened 2026-08-10
feat: the /watchlist page, replacing the nightly HTML report (#147 Part C)

JohnFunkCode:
Part C of [#147](https://github.com/JohnFunkCode/StockPortfolioManager/issues/147) — the `/watchlist` page. This is the report replacement; everything after it in the plan is additive.

The whole shared watchlist, ranked by fundamental score, rendered off the single `GET /api/watchlist/fundamentals` call PR #183 added: one request for the table, not one per row. The server does the ranking and the stale/unscored counts, so the page renders and filters and computes nothing (arch-v2 Rule 8.4).

### Two rules the old report taught us, kept visible in the page

- **Returns arrive as percent, not fractions.** The columns use `formatPercentRaw`; `formatPercent` would multiply by 100 a second time.
- **A market cap is only comparable in one currency.** The native `market_cap` column is deliberately **unsortable**; `market_cap_usd` is the sortable one — ordering mixed currencies ranks exchange rates rather than companies. Every non-USD name is `null` there until an FX source lands ([#185](https://github.com/JohnFunkCode/StockPortfolioManager/issues/185)), and nulls are pinned to the bottom **in both directions**, because a blank at the top of a "biggest first" sort reads as an answer.

### One new route

`PATCH /api/watchlist/{ticker}` takes `{"tags": [...]}` and **replaces** the set rather than merging into it — the UI sends the chip set it is displaying, and a merge could never remove the last tag. An empty array is a legitimate edit, so the service returns the stored list rather than a row count the caller would have to interpret. `PATCH` and not `PUT` because tags are the only editable field; the currency in particular is resolved server-side ([#184](https://github.com/JohnFunkCode/StockPortfolioManager/pull/184)) and must not become client-writable.

Removal stays a **UI-only** action — no `delete` verb was added to `mcp_gateway/rest_client.py`, so no agent can drop symbols off a list the whole team shares. The confirm dialog keeps saying so.

### Sidekick

Ships as GenUI `watchlist_fundamentals`, registered in both `quantcore/services/chat_tools.py` and `frontend/src/chat/componentRegistry.tsx` with an **empty** prop spec: no `owner`, no `symbol`, because the watchlist is one global list ([#83](https://github.com/JohnFunkCode/StockPortfolioManager/issues/83)). Registry parity is asserted in the same PR.

Search, tag filter, and sort model persist in `localStorage`, each parsed in a `try/catch` that falls back rather than blanking the page on corrupt JSON.

### Worth a reviewer's second look

**MUI X DataGrid uses what `getSortComparator` returns verbatim — it does not negate it for `desc`.** So the comparator applies the direction itself and pins nulls *outside* the sign. The obvious `a - b` version left the default `composite_score desc` sort silently ranking **ascending**, with the unscored names on top: no throw, no log, the right arrow in the header. The first assertion of the first page test caught it. Written up in the plan doc's "Notes from PR 4" along with two adjacent traps (DataGrid emits `onSortModelChange` during mount, so localStorage is already populated by the time you look; the `['asc','desc',null]` cycle is why the tests drive an unsorted column).

### Verification

- Backend: `Ran 1379 tests` — **OK (skipped=5)**, against the test database.
- Frontend: **524 tests / 75 files green** (up 50, four new suites); `npx tsc --noEmit` clean.
- Coverage **90.31 lines / 88.32 statements / 88.24 functions / 71.91 branches** against floors 89/87/86/70 — cleared and deliberately **not raised**; `frontend/vitest.config.ts` ratchets once, in Part G.
- OpenAPI surface **107 → 108** (snapshot updated — the gate is route-level, and this adds a route).
- **No schema change**, so no migration.

Docs updated in the same PR: `readme.md` (feature bullet, watchlist files section, PATCH semantics, market-cap sorting, UI pages table), `CLAUDE.md` (route list + the QuantUI page paragraph), and the plan doc's sequencing table and checkpoint log.


JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** `78e2671fbf8e50cceb3c8cb17dd415b4bb081325`
**Verdict:** REQUEST_CHANGES

- **Fully met:** Part C's `/watchlist` page, one-request data flow, persisted search/tag/sort state, currency-safe rendering, tag replacement endpoint, shared-list management, Sidekick registration, and coverage of the new behavior are implemented.
- **Partially met:** The required default ranking is declared as composite score descending then 30-day return descending, but the chosen Community DataGrid does not support the two-column model and warns that it will only use one criterion.
- **Unmet:** None.
- **Unverifiable:** None.

### Code findings
- **frontend/src/components/watchlist/WatchlistPage.tsx:91** — **Warning** — the two-item `DEFAULT_SORT_MODEL` is rejected by the Community DataGrid, so the promised 30-day return tie-breaker is silently not applied.
  **Fix:** Use the Pro multi-sort grid or implement the tie-break in a single supported comparator / pre-sort path, and add a regression test with equal scores and different 30-day returns.
- **frontend/src/components/watchlist/WatchlistPage.tsx:137** — **Warning** — a syntactically valid but malformed `watchlist-tag-filter` value (for example `{}` or `null`) is accepted from localStorage and later crashes filtering because it is not an array.
  **Fix:** Validate `Array.isArray(parsed)` and that every element is a string before accepting persisted state; otherwise fall back to `[]`. Add a regression test for syntactically valid malformed localStorage.

### Verification
- Frontend production build: passed (`tsc -b` and `vite build`).
- Frontend tests: 523 passed, 1 failed in the pre-existing `RotateKeyDialog` test; the run also emitted the new DataGrid warning about the multi-column sort model.
- Backend targeted tests: could not run because the review environment has no `pytest` module installed.
- GitHub checks at review time: secret scan passed; frontend and backend checks were pending.

See the inline locations above for the requested fixes.

*Reviewed by Hermes Agent*

JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** `78e2671fbf8e50cceb3c8cb17dd415b4bb081325`
**Verdict:** REQUEST_CHANGES

**Associated issue:** #147 — Decouple the unused HTML report and S3 upload from main.py, and surface the nightly report in QuantUI

- **Fully met:** The scoped Part C `/watchlist` page is wired through the API and hook layers; it renders the shared watchlist, supports search/tag filtering, tag editing, shared-list removal confirmation, currency-aware values, staleness indicators, navigation, and the GenUI dual registration. The PATCH route replaces tags and preserves the non-tag fields. The relevant backend and frontend behavior is covered by tests.
- **Partially met:** The page's required default ordering is not fully met: the implementation supplies a two-column `sortModel` (`composite_score` then `return_30d`), but the installed community MUI X DataGrid only supports one sort criterion, so the 30-day tie-breaker is ignored.
- **Unmet:** None.
- **Unverifiable:** None.

### Code finding

- **frontend/src/components/watchlist/WatchlistPage.tsx:86-87** — **Severity: Warning** — **What's wrong** — `DEFAULT_SORT_MODEL` contains `composite_score` and `return_30d`, but the community `DataGrid` used here accepts only one sort item, so equal-score rows are not ordered by the promised 30-day-return tie-breaker (the test run emitted `MUI X: The sortModel can only contain a single item...`). **Fix** — either use a DataGrid Pro tier that supports multi-column sorting, or implement the tie-breaker in the row ordering/comparator while keeping the grid's sort model compatible with the community component, and add a regression test with equal scores and different 30-day returns.

### Verification

- `npm run build` — passed (`tsc -b` and Vite production build succeeded).
- `npm test` — 523 passed, 1 failed, 75 test files; `WatchlistPage.test.tsx`'s `PATCHes the whole chip set...` timed out in the full suite, although that same test passed when run alone. The run also consistently emitted the MUI multi-column-sort warning above.
- Targeted backend tests `tests/test_watchlist_service.py tests/test_chat_tools.py` — 59 passed, 6 subtests passed.
- `python3 -m compileall -q api quantcore tests` — passed.
- `git diff --check` — passed.
- Backend repository/API smoke tests were not run because this isolated environment lacks the repository's required `QUANTCORE_TEST_DB_DSN`.

JohnFunkCode:
Both review findings are fixed in f857129. They turned out to be the same shape: something declared that never took effect, with no failure to look at.

### 1. The 30-day tie-break was silently dropped

Confirmed, and the mechanism is worth stating precisely: `disableMultipleColumnsSorting` is a **forced** prop in the Community build (`DataGridForcedPropsKey`), and `sanitizeSortModel` returns `[model[0]]` for anything longer:

```js
// @mui/x-data-grid@7.29.12 — hooks/features/sorting/gridSortingUtils.js
export const sanitizeSortModel = (model, disableMultipleColumnsSorting) => {
  if (disableMultipleColumnsSorting && model.length > 1) { /* warnOnce */ return [model[0]]; }
  return model;
};
```

So the second entry was discarded on every render and the only trace was a dev-mode `warnOnce`. `DEFAULT_SORT_MODEL` is now a single item, and the tie-break moved **inside** the score column's comparator — which is where it has to live, since a comparator is only handed its own column's two values and has to reach the other one through the row:

```ts
function scoreThenReturn(direction: GridSortDirection): GridComparatorFn<number | null> {
  const compare = nullsLast(direction);
  return (a, b, paramsA, paramsB) => {
    const primary = compare(a, b);
    if (primary !== 0) return primary;
    const rowA = paramsA.api.getRow(paramsA.id) as WatchlistFundamentalsRow | null;
    const rowB = paramsB.api.getRow(paramsB.id) as WatchlistFundamentalsRow | null;
    return compare(rowA?.return_30d ?? null, rowB?.return_30d ?? null);
  };
}
```

Reusing `nullsLast` means the tie-break follows the sort direction, while an unknown 30-day return still lands at the bottom — the null pin sits outside the sign, same rule as the primary key.

Two regression tests on three rows tied at 61 with returns 9 / 2 / null: descending expects `BBB, AAA, CCC`, ascending expects `AAA, BBB, CCC`. I checked the descending one **fails** against the previous `getSortComparator: nullsLast`, so it is a guard rather than decoration.

One type note for anyone touching this: `GridSortCellParams.api` is typed `any`, so the generic form `getRow<T>(id)` is a `TS2347` (untyped calls take no type arguments). Cast the result instead.

### 2. Valid JSON of the wrong shape got through

Also confirmed, and it was a page-blanking crash rather than a degrade. The restores were wrapped in `try/catch`, which only covers *unparseable* text — a stored `null` passed the truthy check on the raw string, parsed to `null`, became the filter state, and died at the first `tagFilter.length`. `localStorage` is a user-writable surface, so parse success proves nothing about shape.

Both restores are now shape-validating helpers. The sort model had the same hole one level down: the old `Array.isArray(parsed) && parsed.length > 0` accepted `[{"sort":"asc"}]` and sorted by an undefined field, so items are validated too.

Nine new `it.each` cases across the two keys: `null`, `{}`, a bare string, a mixed `["retail", 7]`, an empty array, a fieldless item, and an unknown direction — each asserting the page renders all three rows on the default sort rather than falling over.

### Verification

535 frontend tests / 75 files green (up 11); `npx tsc --noEmit` clean; coverage **90.34 lines / 88.35 statements / 88.28 functions / 72.02 branches** against floors 89/87/86/70 — cleared and still deliberately not raised, since Part G ratchets them once. No backend change, so the OpenAPI surface and the schema are untouched.

Both traps are written up in the plan doc's "Notes from PR 4" alongside the original comparator gotcha, since the general lessons outlive this page: a grid can decline to honour a plausible declaration, and `JSON.parse` succeeding says nothing about what came back.

JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** `3c36a4784548f587ae8b61533ac2db0e73725503`
**Verdict:** REQUEST_CHANGES

**Associated issue:** #147 — Decouple the unused HTML report and S3 upload from main.py, and surface the nightly report in QuantUI

- **Fully met:** The scoped Part C `/watchlist` page is wired through the API and hook layers; it renders the shared watchlist, supports search/tag filtering, tag editing, shared-list removal confirmation, currency-aware values, staleness indicators, navigation, and GenUI dual registration. The PATCH route replaces tags and preserves non-tag fields. The relevant backend and frontend behavior is covered by tests.
- **Partially met:** The required default ordering is not fully met: the implementation supplies a two-column sort model (`composite_score` then `return_30d`), but the installed community MUI X DataGrid only supports one sort criterion, so the 30-day tie-breaker is ignored.
- **Unmet:** None.
- **Unverifiable:** None.

### Code findings

- **frontend/src/components/watchlist/WatchlistPage.tsx:85-87** — **Severity: Warning** — **What's wrong** — `DEFAULT_SORT_MODEL` contains `composite_score` and `return_30d`, but the community `DataGrid` accepts only one sort item, so equal-score rows are not ordered by the promised 30-day-return tie-breaker; the focused test run emitted the component's warning confirming the model is rejected. **Fix** — either use a DataGrid Pro tier that supports multi-column sorting, or implement the tie-breaker in a single supported comparator/pre-sort path while keeping the grid sort model compatible, and add a regression test with equal scores and different 30-day returns.

- **frontend/src/components/watchlist/WatchlistPage.tsx:128-132** — **Severity: Warning** — **What's wrong** — a syntactically valid but malformed `watchlist-tag-filter` value such as `{}` or `null` is returned directly from `JSON.parse` and later causes `tagFilter.length`/`tagFilter.some` filtering and the multi-select value handling to fail instead of falling back. **Fix** — validate `Array.isArray(parsed)` and that every element is a string before accepting persisted state; otherwise use `[]`, with a regression test for valid-but-malformed localStorage JSON.

- **quantcore/repositories/watchlist_repository.py:202-206** — **Severity: Warning** — **What's wrong** — the required diff-coverage gate fails on the newly added `set_currency` transaction path (only 81% of changed lines are covered, below the repository's 85% floor), leaving commit, rollback, and return behavior untested. **Fix** — add focused tests that exercise successful currency updates and the database-error rollback path (or otherwise cover lines 202-206) until the changed-line coverage reaches at least 85%; the CI gate must pass.

### Verification

- `npm ci --ignore-scripts` — passed; installed 322 packages (npm reported 10 dependency audit vulnerabilities: 1 low, 1 moderate, 8 high).
- Focused frontend tests (`WatchlistPage`, `WatchlistFundamentalsCard`, `watchlist` API, and `useWatchlist`) — **50 passed across 4 files**. The run emitted the MUI X warning that the community grid rejects the two-item sort model.
- `npm run build` — passed (`tsc -b` and Vite production build succeeded).
- Backend targeted tests could not run in this isolated environment because `QUANTCORE_TEST_DB_DSN` is not configured; after installing pytest, collection stopped with the repository's explicit `assert_not_production()` guard for that missing DSN.
- `python -m compileall -q api quantcore tests` — passed.
- `git diff --check` — passed.
- GitHub checks for this head: frontend tests/typecheck/coverage passed; backend `Tests + wrapper smoke + OpenAPI surface` failed its required diff-coverage gate at **81%** (37 changed lines, 7 missing), specifically `quantcore/repositories/watchlist_repository.py` lines 202-206; secret scan passed. Deploy-related checks were skipped.

*Reviewed by Hermes Agent*

PR #188, opened 2026-08-11
feat: the /fundamentals page and scoped fundamentals reads (#147 Parts B4 + D)

JohnFunkCode:
PR 5 of the nine-PR sequence in [`docs/proposals/legacy-report-retirement-plan.md`](docs/proposals/legacy-report-retirement-plan.md) (issue #147). Checkpoint row 5 is filled in.

## Part B4 — `scope` on the fundamentals reads

The four collection reads on `FundamentalsService` take `scope="all"|"tracked"` and echo it back in the payload. **"Tracked" is the shared watchlist plus *every* owner's positions** — deliberately the same set `main.py` builds for the nightly options capture and hands to the fundamentals warmer (#126 decision #5). That is what makes the scope honest in both directions: it never shows a symbol the warmer isn't refreshing, and never hides one the warmer pays for every night.

**`_apply_scope` runs immediately after `get_all_latest()` and before any ranking.** The cache holds every symbol anyone has ever asked about, so the top N of the cache is not the top N of the roster — filtering after ranking would quietly serve a shorter list of the wrong names. On `score-changes` it also cuts real work, since the loop below it runs a history query per surviving symbol.

**Deviation from the plan's wording, worth a reviewer's attention.** Part B4 says to inject the watchlist and portfolio services into `FundamentalsService`. That cannot be done: `WatchlistService` composes `FundamentalsService` (Part B3), so naming them in its constructor is a construction cycle in the registry. What ships instead is a late-bound `tracked_symbols: Callable[[], Iterable[str]] | None` closure wired in `registry.py`, with the two constructions hoisted above the `Services(...)` literal so it resolves to the live instances. It is optional because most construction sites — tests, scripts, the MCP tools — only ever ask for `all`; asking for `tracked` without a provider raises rather than silently returning everything.

Late binding is also what would let a rename or a re-order fail **silently** until a request arrived, so new `tests/test_registry_wiring.py` calls the closure, asserts it equals the union of the two live reads (composition, not a count — prod's positions table is deliberately empty), runs `scope="tracked"` end to end, and pins that `get_services()` hands back the same instances it closed over.

At the route boundary the parameter is `Scope = Literal["all","tracked"]`, so an unrecognized value is a **422 from FastAPI** rather than a service `ValueError` surfacing as a 500. The **OpenAPI surface snapshot does not move** (108 routes) — the gate is route-level, so query parameters are invisible to it, same as PR 2's response fields.

`cache-stats` deliberately takes **no** scope: it describes the cache itself, and scoping it would answer a different question than "how fresh is the data behind this page".

## Part D — the `/fundamentals` page

Five panels over five cache-backed reads — `TopFundamentalsPanel`, `SectorBreakdownPanel`, `ScoreChangesPanel`, `UpcomingEarningsStrip`, `CacheFreshnessBanner` — each owning its own loading and error state with **no page-level gate**, so one slow or failing query can't blank the other four. Tests pin exactly that for a failed `score-changes` and a failed `cache-stats`. This is not hypothetical: score-changes is genuinely the slow one (a history query per surviving symbol) and renders last.

Four panels fetch with `scope=tracked`. The banner is scope-less and fails **silently** — it is a freshness annotation, and an error alert about the annotation is noisier than the missing chip it replaces. It also never paints the database identifier, which a test asserts: the never-log policy covers responses, and `cache_stats()` leaking the DSN is the bug `tests/test_dsn_redaction.py` exists for.

Two sidekick cards ship as GenUI `fundamentals_top` / `fundamentals_score_changes`, both with **empty prop specs** in both registries (whole-universe views — there is neither a symbol nor an owner to name), and both are the panels' own `variant="rail"` sharing one hook and one react-query key rather than second copies. `FundamentalsCards.test.tsx` asserts the rail shape precisely so a future copy-paste shows up as a failure.

## A date bug the preview surfaced

The earnings strip rendered `Aug 9, 2026 · today` for an API value of `2026-08-10`. `formatDate` handed the bare `YYYY-MM-DD` to `new Date()`, which reads a date-only string as **UTC midnight** — one day early for every viewer west of Greenwich, all day. It parses component-wise into local time now; strings carrying a time (`…T12:00:00Z`) are real instants and keep their timezone handling.

Two things to keep straight: it is a **shared-helper** fix, not a page-local one, so `/watchlist`'s `prices <as_of>` chip and both plans pages were a day behind too; and the existing test passed through the whole defect because it asserted `/Jul/` and `/2026/` on a date whose month and year survive the shift. The new case asserts the exact string.

## Verification

Driven against the **test** database (`127.0.0.1:5434`, 228 watchlist rows) through the running UI, not just unit tests:

- `scope=tracked` → 229 in cache; `scope=all` → 233. The filter demonstrably bites.
- `?scope=bogus` → **422 Unprocessable Content**.
- All four data panels request `scope=tracked`; `cache-stats` carries none.
- Page renders: `227 scored / 229 tracked`, `13 movers / 3 unmeasured`, `12 sectors · 229 tracked`, `2 scheduled / 195 stale hidden`, and post-fix `RPD — Aug 10, 2026 · today`.

**585 frontend tests** green (up 50, five new suites), floors cleared at 90.65 lines / 88.7 statements / 88.7 functions / 72.88 branches and **not raised** (mitigation 2 — Part G ratchets them once). **Backend 1404 OK** (skipped=5). No schema change, so the three-files rule is N/A.


JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** `ed824360d62fedd5300d9bc39d23d6eb0adb0968`
**Verdict:** APPROVE

### Code findings
No Critical, Warning, or Suggestion findings. The scoped fundamentals reads filter before ranking, validate `scope` at the FastAPI boundary, and use the late-bound tracked-symbol provider consistently. The `/fundamentals` page keeps loading/error state independent per panel, avoids exposing the database identifier, and shares the rail/page components and query behavior as intended.

Associated issue: #147 — Decouple the unused HTML report and S3 upload from main.py, and surface the nightly report in QuantUI.

- **Fully met:** Part B4's `all|tracked` scope plumbing for the four collection reads, including the shared watchlist plus every owner's positions and pre-ranking filtering; Part D's `/fundamentals` page with the five cache-backed panels, independent panel failure handling, tracked-universe reads, cache freshness display, and the two no-props GenUI sidekick cards.
- **Partially met:** None for the objectives implemented by this PR. The parent issue remains open by design for the other PRs in the nine-PR sequence.
- **Unmet:** None identified for this PR's stated Parts B4 + D scope.
- **Unverifiable:** None based on the available repository CI checks and diff.

### Verification
- Frontend CI: pass — 585 tests, typecheck, and coverage job.
- Backend CI: pass — Tests + wrapper smoke + OpenAPI surface job.
- Secret scan: pass — gitleaks.
- Local frontend tests: 82 files / 585 tests passed.
- Local frontend production build: passed (`tsc -b && vite build`).
- Local backend tests could not run because this review environment lacks `pytest`, `pandas`, and `psycopg2`; the repository backend CI check passed.
- `git diff --check`: passed.

PR #190, opened 2026-08-11
feat: own Harvester plans (#147 Parts H1–H4)

JohnFunkCode:
Item 6 of [`docs/proposals/legacy-report-retirement-plan.md`](docs/proposals/legacy-report-retirement-plan.md) — Parts H1–H4 of issue #147. Harvester plans become owned, the way positions already are.

## What changes

`plan_instances` gains an `owner` column matching `positions.owner`, and the database invariant moves with it: `ux_one_active_plan_per_symbol` is dropped for `ux_one_active_plan_per_owner_symbol`, so two owners may each run a ladder on the same ticker. This is **behaviour-preserving for the single-owner world** — every existing row is John's, which is exactly what the `DEFAULT 'john'` backfills.

**Isolation is enforced in SQL, never in a route.** Every statement carries an `owner` predicate; `plan_rungs` and `plan_alerts` have no `owner` column of their own and inherit it by JOIN to `plan_instances`. So another owner's plan reads as `None` or mutates zero rows, and the routes turn that into a **404** — the same answer as an id that never existed, which is what keeps the endpoint from leaking which ids are taken.

`owner` is a **required keyword-only argument** on all 14 public repository methods and every service passthrough. A caller that forgets it is a `TypeError` at the call site, not a silent read of John's ladders. The one deliberate exception is `_ensure_next_rung_alert`, which is private and only ever reached with an `instance_id` already resolved through an owner-scoped query — re-checking there would only hide a caller that skipped that.

Routes resolve the owner from the authenticated principal via `Depends(require_owner)`. There is **no `?owner=`** on the plans, rungs, dashboard, or symbol-list routes, because there is no legitimate cross-user read path here. `GET /api/symbols/{ticker}/price` reads nothing owned and deliberately takes no owner at all.

`CLOSED` joins the status vocabulary alongside `ACTIVE` and `SUPERSEDED`. The two are distinct: `SUPERSEDED` means "replaced by a newer ladder"; `CLOSED` will mean "the holding is gone", which Part H8 (item 7) puts to work.

## The schema gotcha worth knowing

`_SCHEMA` needed an explicit `ALTER TABLE plan_instances ADD COLUMN IF NOT EXISTS owner …` in addition to the column in the `CREATE TABLE`. `CREATE TABLE IF NOT EXISTS` is a no-op against a database that predates the column, so the new unique index built on `owner` failed with `UndefinedColumn` on startup. Precedent for the fix is the `positions` block a few lines above it.

The `DROP INDEX` / `CREATE UNIQUE INDEX` pair must stay **adjacent and in that order**: under `QUANTCORE_SCHEMA_MODE=create` against a live database there is a window between them with no uniqueness constraint on active plans at all, and anything inserted between them widens it. Both the migration and `_SCHEMA` carry that as a comment.

## Three-files rule

| File | |
|---|---|
| `db/migrations/V7__plan_owner.sql` | new version, fully `IF [NOT] EXISTS` |
| `_SCHEMA` in `quantcore/db.py` | column + the ALTER + the index swap |
| `db/schema_snapshot.json` | regenerated with `scripts/check_schema_snapshot.py --update` |

## Verification

- **Backend: 1422 tests, `OK (skipped=5)`** against the test database.
- Frontend: 82 files / 586 tests pass. Coverage 90.65 lines / 88.7 statements / 88.7 functions / 72.88 branches, clearing the floors — **not raised** (Part G ratchets them once, in PR 8).
- OpenAPI surface unchanged at **108 routes** — the snapshot is route-level, and this adds only dependencies and response fields.
- New `HarvesterOwnerIsolationTest` (8 cases) covers two owners active on the same symbol, cross-owner reads coming back empty, cross-owner mutations changing nothing, and scan / next-actions / dashboard / symbol-list / purge all staying scoped.

## Deploy order

The test database is **already migrated** (V7 applied 2026-08-10; `scripts/schema_check.py` reports 22 tables, 0 missing, 0 extra, and the old index is gone). **Prod still needs `./scripts/flyway.sh --prod migrate` before `prod-rollout.yml` is dispatched** — migrations are load-bearing now, so the image must not land ahead of the schema.


JohnFunkCode (commented):
## Hermes Agent PR Review

**Reviewed head:** `de229ba79e9f3c55936852b196ebb33a634f0229`
**Verdict:** COMMENT

### Code findings

- **quantcore/repositories/harvester_repository.py:94-104; quantcore/repositories/harvester_repository.py:152-158; quantcore/services/harvester.py:162-165 and 192-212** — **Suggestion** — `SQL_GET_NEXT_PENDING_RUNG` and `SQL_GET_ACTIVE_ALERT_FOR_RUNG` do not carry an owner predicate or accept an owner parameter, even though H3 requires every plan/rung/alert query to scope through `plan_instances.owner`; the current callers happen to derive globally unique instance/rung IDs from owner-scoped queries, but the helpers are not independently owner-safe and can silently regress if reused with an unscoped identifier. **Fix:** join `plan_instances` in both queries, add `pi.owner = :owner`, and pass `owner` at every service/repository call site (including the scan path).

### Associated issue/objective assessment

Associated issue: **#147 — Decouple the unused HTML report and S3 upload from main.py, and surface the nightly report in QuantUI**.

The PR explicitly implements the H1-H4 Harvester-plan slice referenced in its body, rather than the issue's entire report-retirement/UI objective:

- **Fully met:** H1 owner column and per-owner unique active-plan index, including V7 and the mirrored schema; H2 `CLOSED` status vocabulary across API/schema/frontend; H4 authenticated owner resolution at routes and notifier edge.
- **Partially met:** H3 SQL-layer owner isolation. The main CRUD and aggregate queries are owner-scoped and the added isolation tests cover the key paths, but the helper queries noted above are not owner-predicate-complete.
- **Unmet:** The parent issue's broader report decoupling and QuantUI report-surfacing work, as well as later H5-H8 plan/holding integration, are not included in this H1-H4 PR.
- **Unverifiable:** None for the implemented H1-H4 behavior.

### Verification

- `git diff --check`: passed.
- Python `compileall` for `api`, `quantcore`, `notifier.py`, and `tests`: passed.
- Local `python -m unittest tests.test_harvester_service -v`: could not run because `QUANTCORE_DB_DSN` / `QUANTCORE_TEST_DB_DSN` is not configured in this environment.
- GitHub checks for this head: backend/wrapper smoke/OpenAPI **success**; frontend tests/typecheck/coverage **success**; gitleaks **success**. Deploy and credential checks were **skipped**.

**COMMENT**

JohnFunkCode:
Fixed in 38a64f4. Taking the suggestion as written — you had the better argument, and the reasoning I'd committed was wrong in a way worth spelling out.

**Confirming the no-defect part first.** All four call sites of `SQL_GET_NEXT_PENDING_RUNG` derive their `instance_id` from `SQL_LIST_ACTIVE_PLANS` (owner-scoped) or, in `record_execution`, from `SQL_GET_RUNG_INSTANCE` (owner-scoped); the single `SQL_GET_ACTIVE_ALERT_FOR_RUNG` call takes its `rung_id` from the row the scoped query just returned. So nothing was leaking or cross-writing today. This is defense in depth, exactly as you labelled it.

**Why the exception was wrong anyway.** I'd documented `_ensure_next_rung_alert` as deliberately unscoped, on the grounds that re-checking would "only hide a caller that skipped scoping." That treats the required keyword and the SQL predicate as alternatives, when they answer different failures:

- the **required keyword** catches a caller that *forgot* to scope — a `TypeError` at the call site;
- the **predicate** bounds a caller that scopes *incorrectly* — passing a well-formed `instance_id` that isn't theirs, which no signature can catch.

And the two outcomes aren't symmetric. Without the predicate, a mismatched `instance_id` writes an alert onto another owner's ladder. With it, the lookup returns nothing and the call no-ops — the same shape every other cross-owner access in this repository already takes, and the shape the routes turn into a 404. Containing a cross-owner write beats preserving a louder symptom for a bug class the keyword already covers. Your phrase "not independently owner-safe" is the right frame; I'd made the guarantee a property of the callers rather than of the query.

**What changed:**

- `SQL_GET_NEXT_PENDING_RUNG` gains `AND pi.owner = :owner` — it already joined `plan_instances`, so this is one line.
- `SQL_GET_ACTIVE_ALERT_FOR_RUNG` now joins `plan_instances` through `alerts.instance_id` and filters on owner.
- `_ensure_next_rung_alert(instance_id, *, owner)`. Three of the four call sites had `owner` in scope already; only `build_plan` needed a keyword added.
- The scan path (`scan_and_fire_alerts`), `get_next_actions`, and `symbols_at_harvest_points` all pass `owner` now, as you asked.

**One thing I added beyond the suggestion.** A test that calls `_ensure_next_rung_alert` directly with a foreign owner and asserts no alert is written — `test_ensure_next_rung_alert_writes_nothing_for_another_owner`. Since every caller reaches the helper already scoped, a route-level test can't distinguish the predicate working from the callers happening to be correct, which is the regression you were pointing at. Asserting on the helper means the property survives a caller being rewritten.

`CLAUDE.md` and the checkpoint log in the plan doc are updated too — both recorded the exception as deliberate, so leaving them would have re-argued this back in on a later PR.

**Verification:** all three harvester suites green — `test_harvester_repository_db` 29 tests OK (1410s), `test_harvester_service` + `test_harvester_fetch_seam` 14 tests OK. No schema change, so the applied V7 migration is unaffected.

PR #191, opened 2026-08-11
fix: keep the test suite importable under the lean CI install

JohnFunkCode:
The prod promotion of `3a8989b` failed its test gate
([run 31509255240](https://github.com/JohnFunkCode/StockPortfolioManager/actions/runs/31509255240/job/93838614215)).
Nothing was wrong in prod — the suite could not import:

```
tests/test_generate_portfolio_report.py:20: import scripts.generate_portfolio_report
scripts/generate_portfolio_report.py:44:    import matplotlib.pyplot as plt
ModuleNotFoundError: No module named 'matplotlib'
```

## Why it only fails here

The two CI jobs install different dependency sets, deliberately:

| Workflow | Installs | Has matplotlib |
|---|---|---|
| `deploy.yml` (every PR + merge to `main`) | `requirements-dev.txt` (base + report + coverage) | yes |
| `prod-rollout.yml` (promotion) | `requirements-base.txt` — the lean set the containers carry | no |

`tests/test_generate_portfolio_report.py` arrived with the report-script
extraction (`4d224b3`, #147 Part F) and imports the script at module level. It
has therefore passed every PR and every merge, and could only ever fail at a
promotion. This was the first prod rollout since it landed.

## The fix, and the two it isn't

The report script is the *legitimate* home for a module-level matplotlib import
— it runs on the Pi, never in a container, which is exactly what
`requirements-report.txt` exists to say. So:

- **Not** moving matplotlib/jinja2/boto3 into function bodies. That hides the
  dependency the split exists to make visible, and buys nothing: the script's
  requirements do not change.
- **Not** installing `requirements-dev.txt` in `prod-rollout.yml`. That would
  delete the only place the lean set is exercised at all.

Instead the **test module** guards its own import and skips, tolerating those
three packages and nothing else — a genuinely broken import in the script still
errors the run. `CLAUDE.md` records the gotcha, because the disagreement between
the two jobs is invisible until a promotion fails.

## Verification

| Scenario | Result |
|---|---|
| Simulated lean install, via `unittest discover` (the path CI takes) | `OK (skipped=1)`, as a `ModuleSkipped` module |
| `numpy` blocked instead of the report deps | still errors — the guard does not over-catch |
| A real missing `psycopg2` (wrong interpreter) | still errored loudly |
| Report deps present (`.venv`) | 9 tests, `OK` |

`grep` confirms this is the only test module in the suite that reaches a
report-only dependency.

## After merge

`deploy.yml` re-rolls test automatically; then re-dispatch `prod-rollout.yml` on
the new SHA. Prod's schema is already at `V7` (migrated 2026-08-11), so nothing
else is pending.

Follow-up worth its own issue: `deploy.yml` never exercises the lean install, so
this whole class of failure can only surface during a promotion. An import-only
smoke on `requirements-base.txt` at PR time would fix that.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,9 +1,10 @@
 # Phase 3 Step 10 — CI/CD for the QuantCore platform (architectural standard v2 §10).
 #
-# PRs + pushes run the gate (tests + wrapper smoke + OpenAPI surface diff) plus a
-# gitleaks secret scan (BYOK packet 8a — a pasted API key or PEM in a diff fails CI
-# before it ever lands; allowlist in .gitleaks.toml). Pushes to main additionally build
-# the 5 images (Cloud Build, native amd64) and roll them out to
+# PRs + pushes run the gate (tests + wrapper smoke + OpenAPI surface diff), the lean
+# install import smoke (requirements-base.txt — the set the containers ship, which the
+# gate does NOT install), and a gitleaks secret scan (BYOK packet 8a — a pasted API key
+# or PEM in a diff fails CI before it ever lands; allowlist in .gitleaks.toml).
+# Pushes to main additionally build the 5 images (Cloud Build, native amd64) and roll them out to
 # the Cloud Run services + the report Job in the TEST project. Deploys carry ONLY the
 # image tag — every service's env/secrets/Cloud-SQL binding is preserved from its last
 # manual deploy (Steps 8–9), so CI never needs the DB DSN or JWT secret.
@@ -114,6 +115,48 @@ jobs:
       - name: Schema snapshot up to date
         run: python scripts/check_schema_snapshot.py
 
+  # The gate above installs requirements-dev.txt (base + report), but the
+  # containers — and prod-rollout.yml's own gate — run on requirements-base.txt
+  # alone. Without this job that difference is invisible until a promotion:
+  # a module importing matplotlib/jinja2/boto3 passes every PR and fails at the
+  # prod gate, which is exactly what happened on 2026-08-10. Separate job rather
+  # than a step, because the two dependency sets cannot share a Python
+  # environment. Import-only — the gate already runs the tests.
+  lean-import:
+    name: Lean install import smoke (requirements-base.txt)
+    runs-on: ubuntu-latest
+    services:
+      # api/main.py builds the app at import time, which calls ensure_schema() —
+      # so even an import-only smoke needs a database to reach.
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_USER: quantcore
+          POSTGRES_PASSWORD: quantcore
+          POSTGRES_DB: quantcore
+        ports: ["5432:5432"]
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+    env:
+      QUANTCORE_DB_DSN: postgresql://quantcore:quantcore@localhost:5432/quantcore
+      AUTH_DISABLED: "1"
+      PYTHONPATH: .
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+          cache: pip
+      - name: Install lean deps (the set the containers ship)
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements-base.txt
+      - name: Import smoke
+        run: python scripts/ci_lean_import_smoke.py
+
   frontend-gate:
     name: Frontend tests + typecheck + coverage
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


JohnFunkCode:
## Hermes Agent PR Review

**PR:** JohnFunkCode/StockPortfolioManager #191 — fix: keep the test suite importable under the lean CI install
**Author:** JohnFunkCode
**Reviewed head:** `a9196d6c138959381fd1c6cbefe0ea18743c3cdf`
**Verdict:** REQUEST_CHANGES
**Job marker:** `hermes-pr-review a9196d6c138959381fd1c6cbefe0ea18743c3cdf`

**Associated issue:** #147 — Decouple the unused HTML report and S3 upload from main.py, and surface the nightly report in QuantUI. The PR body references #147 Part F, but does not state that this PR closes the issue.

- **Fully met:** The PR's scoped objective of adding a lean-install import check and keeping the report-only test import narrow is supported by the changed files and the successful lean-import job.
- **Partially met:** None identified for the scoped deliverable.
- **Unmet:** None identified for the scoped deliverable; the broader implementation objectives in #147 are outside this PR's stated scope.
- **Unverifiable:** The broader #147 objective of decoupling the report and surfacing it in QuantUI is not established by this CI-focused patch and should be assessed against the implementing PRs.

### Finding
- **.github/workflows/deploy.yml:144** — exact location
- **Severity** — Warning
- **What's wrong** — The new `lean-import` job sets `QUANTCORE_DB_DSN` to `postgresql://quantcore:***@localhost:5432/quantcore`, while its Postgres service sets `POSTGRES_PASSWORD: quantcore`, so the job's API import can fail authentication instead of exercising the intended lean-install smoke against the service database.
- **Fix** — Use a DSN whose password matches the service (for example `postgresql://quantcore:***@localhost:5432/quantcore`, URL-encoding it if needed), or set the service password to match the DSN; add a small connection/startup assertion so a credential mismatch cannot make this gate misleading.

### Verification
- `python3 -m py_compile scripts/ci_lean_import_smoke.py tests/test_generate_portfolio_report.py` — passed.
- `git diff --check` — passed.
- GitHub Actions run [31511612644](https://github.com/JohnFunkCode/StockPortfolioManager/actions/runs/31511612644): lean-import completed successfully; secret scan succeeded; frontend gate and tests/wrapper/OpenAPI job were still `in_progress` at review time, so those are not treated as passed.

JohnFunkCode:
## Re: review finding on `.github/workflows/deploy.yml:144` (credential mismatch)

**Not reproducible — the passwords already match.** 

The suggested fix is
character-for-character what line 144 already is. The same service/DSN pair also appears at lines
57/67 for the pre-existing `gate` job, where it has been green for months, and in
`prod-rollout.yml`'s gate — `lean-import` copied it deliberately so there is one credential pair
across the workflow rather than two to keep in sync.

**The run is the proof.** CI run `31511612644` on `a9196d6` reports
`Lean install import smoke (requirements-base.txt): success`, with
`lean import smoke: 86 modules, 0 failed, 1 skipped`. `api.main` is in that module list, and
`api/main.py` builds the app at import time (`app = create_app()` → `ensure_schema()` →
`psycopg2.connect`), so the job cannot go green without a completed authenticated connection to the
service database. An auth failure here is loud, not silent: when the socket was blocked during local
testing, the same script returned 8 failures and exit 1.

**On "add a small connection/startup assertion so a credential mismatch cannot make this gate
misleading":** importing `api.main` *is* that assertion — it connects and runs the schema check
before the smoke can report success. A second one would assert the same thing twice.

There is a real point buried in it, though, and I've taken that part. A credential or connectivity
failure could never make the gate *pass*, but it would have been **misdiagnosed**: every failure
printed the same trailing hint blaming `requirements-base.txt`, when a DB problem surfaces as an
import failure without being one. `scripts/ci_lean_import_smoke.py` now branches on whether any
failure was an `ImportError` — the lean-install hint only prints when something actually failed on a
missing module, and otherwise it points at the environment:

> Nothing above failed on a missing module, so this is not a lean-install problem — read the
> tracebacks. A connection error is the environment, not the dependency set: importing api.main runs
> create_app() -> ensure_schema(), which needs the job's Postgres service reachable at
> QUANTCORE_DB_DSN.

Verified locally against all four cases: full deps (`0 failed, 0 skipped`), report packages blocked
(`0 failed, 1 skipped`), a blocked lean-set module (`24 failed`, lean hint, exit 1), and an
unreachable database (`8 failed`, environment hint, exit 1).

Recommend dismissing the credential half of the finding as inaccurate.

JohnFunkCode:
Correction to the timing in my comment above: this PR merged 41 seconds before the diagnostic-message change was pushed, so it is **not** in `main` — the branch ended up one commit ahead of the merge, and no CI ran on it (the `pull_request` trigger was already gone).

Carried into #192, from the same branch. Nothing here is wrong as merged — the `lean-import` job and the smoke work; a database failure would still fail the job red. Only the trailing hint it prints on that failure is misdiagnosed, and #192 is the fix.

PR #199, opened 2026-08-11
feat: a harvest plan may only exist while the owner holds the shares (#147 H5-H8)

JohnFunkCode:
Sequencing item 7 of [`docs/proposals/legacy-report-retirement-plan.md`](docs/proposals/legacy-report-retirement-plan.md) — Parts H5–H8 of #147.

## The invariant

A harvest ladder sells into strength. One running on a symbol nobody holds fires alerts that can
never be executed. So: **a plan may only exist while its owner holds the shares**, enforced at both
ends of the position's life.

**Entry (H5).** `HarvesterService.build_plan` refuses a symbol with no `OPEN` lot **before** the
yfinance fetch, raising `RuntimeError` — which `POST /api/plans` already maps to **422**. That is the
right code: the request is well-formed, the portfolio just doesn't support it. `CreatePlanDialog`
narrows its symbol picker to the caller's holdings, but the picker is the courtesy and the 422 is the
rule.

**Exit (H5).** `PortfolioService._close_plan_if_flat` closes the owner's ACTIVE plan to **`CLOSED`**
— not `SUPERSEDED`, since nothing replaced it — once no `OPEN` lot remains, on the three paths that
can empty a position: `close_lot`, `delete_lot`, `remove_position`. It runs **after** the write has
committed, on its own connection, inside a `try/except` that only logs: a harvester outage must not
fail a sale that already landed.

A **partial** close leaves lots open and therefore leaves the plan ACTIVE. That off-by-one is the
difference between finishing a harvest and silently killing a live ladder, so it has its own test.

## Why repositories in both directions

`HarvesterService` takes `portfolio_repository`; `PortfolioService` takes `harvester_repository`.
Injecting the *services* in both directions would close a construction cycle in `registry.py`. Both
are optional (`None`) so a unit test can build either stack alone.

## The backstop (H7) and the chip (H6)

The exit seam is eventually consistent by design — it fires after a commit, in a swallow-and-log
block — so `GET /api/plans` carries `in_portfolio` per plan and `/plans` flags an ACTIVE row whose
shares are gone.

`None` where the holdings could not be read, and the page flags only an explicit `false`.
**"We can't tell" is not "they sold out of it"**, and a UI that conflates the two teaches people to
ignore the flag.

`GET /api/symbols` carries `active_plan_id` for the Portfolio page's Plan chip, looked up **once per
request** (`HarvesterPlanDB.active_plan_ids`) and degrading to `None` rather than failing the table.

## The backfill (H8), and what the pre-flight found

`V8__close_orphan_plans.sql` closes rows that predate the invariant. It is a **data** migration, and
is deliberately **not** mirrored into `_SCHEMA` or `db/schema_snapshot.json` — `init_schema()` runs on
every application startup, and a backfill that re-ran there would close plans built moments earlier by
a request whose position write had not yet landed.

**Applied to test** (schema version 7 → 8). **Prod is still at V7** and must be migrated before
`prod-rollout.yml` is dispatched.

Both environments were pre-flighted read-only:

| | `plan_templates` / `plan_instances` / `plan_rungs` | OPEN positions | Rows `V8` closes |
|---|---|---|---|
| test | 0 / 0 / 0 | 5 (`john`, `thomas`) | **0** |
| prod | 0 / 0 / 0 | 4 — GEV, GLW, TER, ZS, all `john` | **0** |

Which is the thing to be careful about: **a green `--prod migrate` is not evidence the backfill
works.** There are no plan rows anywhere to act on, so it will report zero either way. The unit tests
are the evidence; the migration's value here is prospective, guarding plans built after the invariant
lands.

Running the pre-flight also corrected three things the plan doc asserted, all now fixed in the doc:

- The pre-flight query **disagreed with the migration**: it hardcoded `p.owner = 'john'` where `V8`
  correlates on `p.owner = pi.owner`. Since H1 made plans owned, the hardcoded form reports another
  owner's plan as an orphan that `V8` would correctly leave alone. A pre-flight that disagrees with
  its migration is worse than none — it is read once, believed, and never rechecked.
- The claim *"prod holds exactly one open position — ZS"* was stale; there are four.
- The runbook command didn't work. `.claude/with-test-db.sh` **exports** the DSN rather than passing
  it, and `psql` doesn't read that variable, so the documented `with-test-db.sh psql -f …` fell
  through to the local socket. The `sh -c 'psql "$QUANTCORE_DB_DSN" -f …'` wrapper is load-bearing.

## Review fix: a symbol the picker can't show

`initialSymbol` seeds the field on the first render; the holdings query resolves later. In between,
the MUI select holds a value with no matching `MenuItem` — blank field, plus an out-of-range warning.

The warning is cosmetic; the state it reports can outlive the fetch. Open the dialog on an
`initialSymbol` the caller does not hold — a stale Portfolio row, a position closed in another tab —
and the field stays blank while `symbol` still carries the ticker, so Create stays enabled and
submits a symbol nobody was shown. The user gets a 422 about a security the form never displayed.

So the select's value is now derived from what it can actually show, and submission is gated on the
same value: blank field, disabled button, one story. Both new tests fail against the unfixed
component — one on the warning, one on the live button.

## Verification

| Check | Result |
|---|---|
| Targeted: `test_harvester_repository_db`, `test_harvester_service`, `test_portfolio_service` | **100 tests, OK** (1894.6s) |
| Full backend suite | **1446 tests**, one failure — `test_store_and_read_back`, unrelated to this branch and since fixed on `main` by #198 |
| Frontend | **600 tests**, floors cleared at 90.67 lines / 88.73 statements / 88.54 functions / 73.17 branches |
| OpenAPI surface | **108 routes**, unchanged — the new response fields are additive and the gate is route-level |
| Schema snapshot | **22 tables**, unchanged — `V8` is data, not DDL |

The vitest floors are **not** ratcheted here. Part G (item 8) raises them once, rather than three
concurrent frontend PRs each conflicting on the same four-line `thresholds` block.

## Docs

`CLAUDE.md`'s Harvester section gains the holding invariant, both enforcement points, the
repository-injection rule and its reason, and the `in_portfolio` / `active_plan_id` fields;
`readme.md`'s UI tour gains the Portfolio plan chip; the plan doc gains the item 7 checkpoint row,
the pre-flight results, and the three corrections above.


JohnFunkCode (commented):
## Hermes Agent PR Review

**Reviewed head:** `d6801b6eed1e921f1a925d8831856d3ccbb0bba8`
**Verdict:** COMMENT

### Code findings

- **frontend/src/components/plans/CreatePlanDialog.tsx:20** — **Suggestion** — `useState(initialSymbol ?? '')` assigns the preselected symbol before the asynchronous portfolio query has populated the MUI `select`, producing an out-of-range-value warning (reproduced for `WMT`/`AAPL`) while the dialog is opening.
  **Fix:** Keep the select value empty until the holdings have loaded, or render the preselected symbol only after confirming it is in `symbols`; alternatively render a temporary non-select/loading field until the options exist.

**Associated issue:** #147 — Decouple the unused HTML report and S3 upload from main.py, and surface the nightly report in QuantUI.

This PR explicitly scopes itself to sequencing item 7, Parts H5–H8 of the approved plan referenced by issue #147.

- **Fully met:** H5 holding invariant at plan creation and position exit (including owner scoping, partial-close behavior, and failure isolation); H6 owner-scoped Portfolio Plan chip; H7 explicit `in_portfolio` orphan flag with `None` when holdings cannot be read.
- **Partially met:** H8 migration implementation and owner-correlated pre-flight are present, but the recorded test migration had zero `plan_instances` rows and therefore did not verify that the backfill closes an orphan while preserving another owner's plan.
- **Unmet:** None identified within this PR's H5–H8 scope.
- **Unverifiable:** Production pre-flight/migration impact cannot be verified from this checkout or test database.

### Verification

- Frontend: **598 tests passed** across 82 test files (`npm test -- --run`). The run emitted existing React `act(...)` warnings and the changed dialog emitted MUI out-of-range select warnings for preselected symbols before holdings loaded; the latter is reported above.
- Python changed-area syntax: `python -m compileall -q api quantcore tests` **passed**.
- Targeted Python tests: **could not run** because `pytest` is not installed in the review environment and the unittest suite requires `QUANTCORE_DB_DSN`/the repository test database, which was unavailable.
- Production migration/pre-flight: **not run**; no production database credentials were available.

No Critical or Warning-level code findings were identified. The frontend tests provide meaningful coverage for the new picker, Plan chip, orphan flag, and position-exit behavior; the missing populated migration fixture remains the main verification gap.

PR #201, opened 2026-08-11
feat: group the nav into two menus and retire /symbols (#147 Part G)

JohnFunkCode:
Part G of [issue #147](https://github.com/JohnFunkCode/StockPortfolioManager/issues/147) — item 8 of nine in [`docs/proposals/legacy-report-retirement-plan.md`](docs/proposals/legacy-report-retirement-plan.md). Front-end only; no route, schema, or service changes.

## G1 — `/symbols` is retired from the front end

It was a Harvester-era symbol registry promoted to top-level nav, showing three of its four columns better on `/securities` and the fourth on `/plans`.

**Gone:** `SymbolsPage.tsx` + its test, `symbolsApi.list`, the `useSymbols()` list hook, and the now-orphaned `SymbolInfo` type.

**Kept, deliberately:** `components/symbols/LivePrice.tsx` — it backs `PlanDetailPage` and is registered as the `live_price` sidekick component, so the page goes and the folder stays. Both `/api/symbols*` routes stay too; the list route still carries `active_plan_id` for MCP callers.

**A doc that was wrong before this PR.** CLAUDE.md said `GET /api/symbols` carries `active_plan_id` "for the Portfolio page's Plan chip", which reads as though removing the front-end list hook breaks the chip. It doesn't — the chip reads `/api/portfolio/symbols` (`PortfolioService._active_plan_ids`), and the Harvester-era list route independently carries the same field from its own owner-scoped join. Two routes, two lookups, one sentence conflating them. Corrected here.

## G2 — the nav becomes two dropdown menus

| Menu | Pages |
|---|---|
| **My Positions** | Portfolio, Plans, Harvester |
| **Research** | Securities, Watchlist, Fundamentals, Arbitrage |
| *top-level* | Settings |

- Grouping is keyed on the **label, in first-appearance order** (`buildNavSections`), not on adjacency — so a page appended mid-`routes` joins the existing menu instead of opening a second one with the same name. To get the menu order out of that, `routes` itself is reordered rather than adding a parallel ordered group declaration; that also keeps each detail route adjacent to its parent.
- `buildNavSections` is exported **only** so a test can prove the label-keying: the real array is already sorted, so `navSections` alone cannot tell label-keyed from adjacency-keyed grouping apart.
- The toolbar moves to `frontend/src/components/layout/NavBar.tsx` because each group owns a menu-anchor `useState` and `App.tsx` should not grow one per group. `App.tsx` drops 118 lines of chrome and is now `Layout` + `<Routes>`.
- Active state is **`aria-current`** — `'page'` on a link, plain `true` on a group trigger that is not itself the page. Chosen over asserting computed CSS: a real accessibility signal *and* a stable assertion target.

## Tests and the coverage ratchet

**608 frontend tests pass**; `npx tsc -b --noEmit` clean. New `NavBar.test.tsx` asserts: only the two menus + Settings are reachable until a menu opens, Research reveals all four pages, selection navigates *and* closes the popover, `/watchlist` lights Research while My Positions stays dark, and `/plans/7` keeps its parent lit with `Plans` marked `aria-current="page"`.

The vitest floors are ratcheted **once, here**, as the plan reserved (mitigation 2) rather than every intermediate PR:

| | lines | statements | functions | branches |
|---|---|---|---|---|
| measured | 90.75 | 88.81 | 88.65 | 73.39 |
| floor was | 89 | 87 | 86 | 70 |
| floor now | **90** | **88** | **88** | **72** |

## Two things that cost time

- Clearing the menu anchor in the `MenuItem` `onClick` is load-bearing — without it the popover sits open over the page it just navigated to.
- `testUtils.tsx`'s `Providers` is not enough for this component: `useChat()` throws outside `<ChatProvider>`, so `NavBar.test.tsx` copies `App.test.tsx`'s `AppThemeProvider → ChatProvider → MemoryRouter` stack.

## Docs

readme's Pages table gained a **Menu** column and lost the Symbols row; CLAUDE.md gained the `navigation.tsx` declaration rules (and the `active_plan_id` correction above); the plan's checkpoint log has its item-8 row.


JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** `d0de5a33a248036479bc687e1bd64065d81a8434`
**Verdict:** APPROVE

### Code findings

No Critical, Warning, or Suggestion findings. The Part G changes are internally consistent: the `/symbols` frontend page/list hook are removed while the API routes and `LivePrice` consumer remain, and navigation grouping, active-state semantics, route handling, and menu-close behavior are covered by focused tests.

**Associated issue:** #147 — Decouple the unused HTML report and S3 upload from main.py, and surface the nightly report in QuantUI

The PR explicitly implements Part G of the issue's retirement plan:

- **Fully met:** Part G1 — retire `/symbols` from the frontend, remove its page/list hook/orphaned type, retain the API routes and `LivePrice` consumer, and correct the misleading documentation.
- **Fully met:** Part G2 — group the navigation into My Positions and Research menus, keep Settings top-level, preserve detail-route behavior, and expose active state through `aria-current`.
- **Unverifiable:** The broader issue objectives outside Part G (report decoupling and other retirement-plan parts) are not deliverables of this PR and were not re-reviewed here.

### Verification

- Focused frontend tests: **22 passed** (`NavBar.test.tsx`, `navigation.test.tsx`).
- TypeScript: **passed** (`npx tsc -b --noEmit`).
- Full frontend suite: **607 passed, 1 failed**. The failure is in the unrelated pre-existing `src/components/settings/RotateKeyDialog.test.tsx` test (`onClose` expected once, received zero); no changed file is in that test's path or implementation.
- Remote checks at review time: lean-install import smoke **passed**; secret scan **passed**; frontend tests/typecheck/coverage and backend tests/wrapper smoke/OpenAPI were still **pending**, so they are not treated as passed.

PR #206, opened 2026-08-12
fix: Create Plan died on float(None) when the read window outran the fetch

JohnFunkCode:
## The bug

Create Plan failed with `float() argument must be a string or a real number, not 'NoneType'` — from the Portfolio page and the Plans page alike, since both post to the same route.

## Root cause

`PlanBuildParams.history_window_days` (default 360) is a count of **bars** — the repository reads the history back as `LIMIT history_window_days` rows — but the fetch converted it to **calendar** days (`max(n + 60, 420)`). 420 calendar days buys ~288 sessions, so the oldest ~72 rows of a 360-bar window were never written by the harvester.

They were served instead by the *other* writer of `ohlcv`. `OhlcvRepository` (the prices cache) fetches with `auto_adjust=True`, has no adjusted close to offer, and wrote `adj_close = NULL` — and its `ON CONFLICT … adj_close = EXCLUDED.adj_close` also erased values the harvester had already computed for the same rows.

Measured on the test database, to the day:

| | |
|---|---|
| GLW read window | `2025-03-06 .. 2026-08-11` (360 rows) |
| NULL `adj_close` rows | 72, spanning `2025-03-06 … 2025-06-17` — all the oldest |
| first real `adj_close` | index 72 = `2025-06-18` |
| 420 calendar days ago | `2025-06-18` |

Blast radius: 254 symbols have 1d history, 248 have ≥360 rows, and all 248 had rows older than the 420-day fetch reach. Create Plan was broken for essentially every symbol with cached history. Prod runs the same code with the same nightly warming, so assume the same exposure there.

## The fix — three layers, because the middle one can still fail

1. **Fetch the span in trading days.** `HarvesterService.build_plan` scales `history_window_days` by `365/252` before calling `fetch_history`, keeping the old `+60` slack. 360 bars → 581 calendar days ≈ 398 sessions.
2. **A price refresh may fill `adj_close`, never blank it.** `OhlcvRepository`'s upsert becomes `adj_close = COALESCE(EXCLUDED.adj_close, ohlcv.adj_close)`. A tree-wide grep confirms `HarvesterPlanDB.build_plan` is the only consumer of the column, so nothing depended on the nulling.
3. **When a gap survives anyway, say so loudly.** `HarvesterPlanDB.build_plan` raises a `RuntimeError` naming the symbol, the gap count and the oldest date — which `POST /api/plans` already maps to **422** — rather than substituting `close`. Splicing an unadjusted bar in beside adjusted ones is a fake step that inflates the volatility sizing the whole ladder; a legible refusal beats a plausible wrong ladder.

## Also, adjacent (requested in review of the diagnosis)

The bar insert was ~580 single-row round trips. Now one `conn.executemany`, which pages through `psycopg2.extras.execute_batch` — deliberately *not* folded into one multi-VALUES insert, since a repeated `bar_date` would then fail `cannot affect row a second time` against the `ON CONFLICT`. Through the Cloud SQL auth proxy that is the difference between a minute of Create Plan and a second.

## Verification

- `Ran 91 tests in 408.321s — OK` across `test_harvester_fetch_seam`, `test_harvester_repository_db`, `test_repositories_db`, `test_harvester_service`.
- Four new regression cases: the fetch span covers the read window in business days; a read window reaching unadjusted rows raises rather than crashes; unadjusted rows *outside* the window stay harmless; a price refresh never blanks an adjusted close. The two DB-backed ones live in their own `HarvesterAdjCloseGapTest` sibling class — added to `HarvesterRepoTest` they would have run three times, since two suites subclass it.
- **End to end against the test database via the call that failed:** `POST /api/plans` → **201** for GLW (`instance_id 2554`, 4 ascending rungs 206.95/236.51/275.93/331.12) and TER, in 3.2s including the yfinance fetch. `GET /api/plans` reads both back with `in_portfolio: true`. Post-fix coverage: 553 bars, 398 with a real `adj_close`, so the 360-bar window sits inside it with ~38 bars of slack.
- No MCP surface exposes plans (`grep -rln "plans" fastMCPTest/*.py` is empty), so REST and the UI are the only callers — nothing else to exercise.

## Docs

`CLAUDE.md` gains a block under the unified-database section: `ohlcv` has two writers and they disagree about `adj_close`; a refresh may fill it but never blank it; a window measured in bars must be fetched in trading days.

No schema change — `adj_close` was already nullable — so no migration, no `_SCHEMA` edit, no snapshot update.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -242,6 +242,24 @@ All persistence is consolidated into a single **QuantCore** PostgreSQL database,
 
 All repositories under `quantcore/repositories/` and the REST API (`api/main.py`) use the shared factory instead of managing individual database connections.
 
+**`ohlcv` has two writers, and they disagree about `adj_close`.** `OhlcvRepository` (the prices
+cache) fetches with `auto_adjust=True`, so the adjustment is already baked into `close` and it
+writes `adj_close = NULL`; `HarvesterPlanDB.build_plan` fetches with `auto_adjust=False,
+include_adj_close=True` and writes a real adjusted close for the same `(symbol, '1d', ts)` rows.
+Two rules follow, and both exist because breaking either produced the same crash — `float(None)`
+escaping as a bare `TypeError` into the Create Plan dialog:
+
+- **A price refresh may fill `adj_close`, never blank it.** `OhlcvRepository`'s upsert uses
+  `adj_close = COALESCE(EXCLUDED.adj_close, ohlcv.adj_close)`; with plain `EXCLUDED` a routine
+  symbol lookup silently destroyed every adjusted close the harvester had computed.
+- **A window measured in bars must be fetched in trading days.** `PlanBuildParams.history_window_days`
+  is a row count (`LIMIT history_window_days`), so `HarvesterService.build_plan` scales it by
+  `365/252` before calling `fetch_history`, which takes calendar days. The old `max(n + 60, 420)`
+  covered ~288 sessions of a 360-bar read, and the shortfall was served by whatever the prices cache
+  had left there. When a gap survives anyway, the repository raises a `RuntimeError` naming it
+  (→ 422) rather than substituting `close` — an unadjusted bar beside adjusted ones is a fake step
+  that inflates the volatility sizing the ladder.
+
 **Migrating from a legacy SQLite database:** `scripts/migrate_sqlite_to_postgres.py` performs a one-shot copy of an existing `quantcore.sqlite` file into PostgreSQL — it initializes the schema, migrates all 16 tables in FK-safe order via batched `execute_values()` inserts, resets `SERIAL` sequences, and verifies row counts. Run it with `--sqlite <path>` and `--dsn <postgresql-uri>`.
 
 ### Services Layer (`quantcore/`)
--- a/quantcore/repositories/harvester_repository.py
+++ b/quantcore/repositories/harvester_repository.py
@@ -329,25 +329,34 @@ class HarvesterPlanDB:
         if bars.empty:
             raise RuntimeError(f"No price history supplied for {symbol}")
 
+        # One executemany, not a row-at-a-time loop: conn.executemany() pages
+        # through psycopg2's execute_batch, so ~580 daily bars cost ~2 round
+        # trips instead of 580. Through the Cloud SQL auth proxy (~90ms RTT)
+        # that is the difference between a minute of Create Plan and a second.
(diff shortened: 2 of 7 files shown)


JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** b8b695118ee7f19bef2df270cdefeb3dad26edec
**Verdict:** APPROVE

### Code findings

- **tests/test_harvester_repository_db.py:418** — **Suggestion** — `HarvesterAdjCloseGapTest` subclasses `HarvesterRepoTest`, so unittest inherits and reruns every base test in this new class despite the comment calling it a sibling; make it a standalone `unittest.TestCase` (or move only the shared setup/helpers) so the regression tests do not duplicate the entire repository suite.
  **Fix:** Change the class to inherit from `unittest.TestCase` and provide only the database setup/cleanup and `build` helper it needs, or otherwise use composition rather than inheriting the test cases.

**Associated issue:** None found. GitHub `closingIssuesReferences` was empty, and the PR title/body contained no unambiguous same-repository issue reference.

- **Fully met:** No associated issue objectives.
- **Partially met:** None.
- **Unmet:** None.
- **Unverifiable:** None.

The PR addresses the reported `float(None)` path by fetching a trading-day-sized history, preserving existing adjusted closes during price-cache refreshes, and returning a descriptive runtime error instead of mixing adjusted and unadjusted prices. The new tests provide meaningful protection for the fetch-span, adjusted-close gap, and non-blanking behaviors; the inherited test class is the non-blocking test-quality observation above.

### Verification

- `python3 -m compileall -q quantcore tests` — passed.
- GitHub Actions `Tests + wrapper smoke + OpenAPI surface` — passed: https://github.com/JohnFunkCode/StockPortfolioManager/actions/runs/31552740554/job/93978715406
- GitHub Actions `Frontend tests + typecheck + coverage` — passed: https://github.com/JohnFunkCode/StockPortfolioManager/actions/runs/31552740554/job/93978715490
- GitHub Actions `Lean install import smoke (requirements-base.txt)` — passed: https://github.com/JohnFunkCode/StockPortfolioManager/actions/runs/31552740554/job/93978715528
- Focused local pytest runs were unavailable because the review environment has no `pytest` module; database-backed tests also require the repository's PostgreSQL test environment.
- The `Secret scan (gitleaks)` check failed before scanning because the workflow lacks the required `GITLEAKS_LICENSE`; this is an infrastructure/configuration failure, not evidence of a secret in this PR.

APPROVE

PR #209, opened 2026-08-15
Close the Sidekick tool gap, and fix the docs that hid it (#208)

### LINKED ISSUE #208: Research: Sidekick doesn't have access to all MCP server tools — audit against capabilities-matrix.md and update it
## The observation

Sidekick's tool vocabulary (`quantcore/services/chat_tools.py`) does not include every capability
that exists across the project's MCP servers, and it isn't obvious whether the gap is a deliberate
curation (context/token budget, chat-appropriate scope) or accidental drift as servers and tools
were added over time.

Two concrete data points from a quick pass, neither of which matches
[`docs/capabilities-matrix.md`](docs/capabilities-matrix.md) (last updated 2026-07-20):

- The matrix documents **5** MCP servers (stock-price, company-fundamentals, options-analysis,
  news-sentiment, market-analysis / 51 unique tools). The codebase actually registers **7**
  (`fastMCPTest/*_server.py`) — `portfolio-server` and `arbitrage-server` exist and are wired into
  `.mcp.json` but appear nowhere in the matrix's server count or domain tables.
- The matrix says Sidekick has "7 data tools" (`get_stock_price`, `get_technical_signals`,
  `get_rsi`, `get_macd`, `get_fundamental_score`, `get_news_sentiment`, `price_vertical_spread`).
  `chat_tools.py` actually registers **10** data tools today — those 7 plus `analyze_arbitrage_pair`,
  `scan_arbitrage`, and `list_arbitrage_universe` — plus `show_component`. Coverage of the
  arbitrage domain is partial (`discover_arbitrage_pairs` is not exposed), and `portfolio-server`
  (`get_portfolio`, `list_watchlist`, `add_to_watchlist`, `get_symbol_lots`) has **zero** Sidekick
  exposure — a chat user can't ask Sidekick "what's in my watchlist?" even though the MCP tool
  exists.

So the matrix is stale on both ends of the comparison, and the underlying question — *what should
Sidekick have access to, and why doesn't it* — hasn't actually been answered anywhere.

## Task 1 — Research the gap

- Build a definitive current-state inventory: enumerate every MCP server (`fastMCPTest/*_server.py`
  + `mcp_gateway/`), every tool each one registers, and diff that against
  `BACKEND_COMPONENT_REGISTRY` / the tool list in `quantcore/services/chat_tools.py`.
- For each tool that's on an MCP server but *not* in Sidekick's vocabulary, determine why: was it a
  deliberate exclusion (e.g. too large a payload for chat, not GenUI-renderable, owner-scoped data
  that doesn't fit the chat's auth model, redundant with an existing tool) or was it simply never
  added when the server/tool shipped? Check git history / PR descriptions for the tools in question
  where it helps establish intent.
- Note the arch-v2 constraint this interacts with: anything Sidekick renders as a component needs
  matching registry entries in **both** `chat_tools.py` and
  `frontend/src/chat/componentRegistry.tsx` (Rules 8–9) — a plain data tool doesn't need that, but a
  new `show_component`-rendered one does. Call out which candidate additions are "just a tool" vs.
  "tool + component."
- Produce a recommendation: which currently-missing tools should be added to Sidekick (and roughly
  what that takes), and which should stay excluded with the reason written down so this doesn't get
  re-litigated later.
- Related but distinct: issue #121 (Sidekick ranked-list bug) previously root-caused a *system
  prompt* conflict, not a missing-tool problem — worth a skim so this research doesn't rediscover
  that ground, but it's a separate fix.

## Task 2 — Bring `docs/capabilities-matrix.md` back in line with the codebase

Using the findings from Task 1:

- Correct the MCP server inventory (7 servers, not 5) and re-verify the tool-count-by-server table
  against source rather than carrying forward the 2026-07-20 numbers.
- Correct the "Sidekick chat tools" count/list throughout (the summary table, the Sidekick Chat &
  BYOK domain section, and the "Sidekick partially mitigates Tier 1" note) to match
  `chat_tools.py` as it actually stands after Task 1's changes (if any) land.
- Add `portfolio-server` and `arbitrage-server` to the "Overview: The Six Surfaces" / domain tables
  wherever they're currently absent.
- Bump "Last Updated" and the document version, and note in the changelog line what prompted the
  refresh (this issue).

## Why this matters

The matrix exists specifically to keep MCP/REST/WebUI/Sidekick surface parity visible and
actionable (see its own headline section on WebUI gaps). A stale inventory defeats that purpose
quietly — the next reader trusts a server/tool count that's already wrong.

JohnFunkCode:
Closes #208.

## The question, and the actual answer

#208 asked why Sidekick doesn't have access to all the MCP servers' tools. The audit gives a three-part answer, and only one part is what the title suspected.

**1. The curation is deliberate policy, not drift.** `architectural-standard-v2` §5.5 says exposing an endpoint as a tool is a decision, and the arbitrage commit (`c5df10f`) shows that reasoning applied in practice. Working as designed.

**2. `SYSTEM_PROMPT` was stale, and this is the likeliest real cause of the symptom.** It hard-coded 4 of the 15 registered components and never mentioned the arbitrage, portfolio, or fundamentals tools at all. A model does not reach for a tool its own prompt never names.

**3. Two gaps were genuinely accidental.** `arbitrage_discovery` was registered as a component on *both* sides of the registry with no tool able to feed it, and `portfolio-server` had zero chat coverage — Sidekick could render a portfolio card but couldn't reason over the numbers in prose.

## What's here

Six commits, ordered for review: owner plumbing → `chat.py` → `chat_tools.py` → `registry.py` → tests → docs. Splitting it would either duplicate the owner-dispatch plumbing across two PRs or land the docs against a tool count that isn't real yet — which is the exact failure this issue is about.

### Three worth arguing about

**Owner resolution degrades per tool, not per request.** `require_owner` 403s the whole request on an unmapped identity. That's wrong for chat: a turn serves many tools that need no identity, and failing all of them over unprovisioned portfolio identity is a bad trade — while defaulting to `"john"` would be a silent cross-user read the moment a second identity exists. So `resolve_owner_or_none` returns `None`, and the dispatch loop fails that *one* tool with a recoverable error the model can explain. It never auto-provisions, and logs a distinct event so the soft path stays separable from the hard 403 in monitoring.

**Discovery is now stricter than its arbitrage siblings — deliberately.** `MAX_DISCOVER_SYMBOLS` (25) / `MAX_DISCOVER_REFERENCES` (10) lived only in `api/routers/arbitrage.py`. Sidekick dispatches **in-process** and never passes through that route, so the cap would simply not have existed for chat. They now live on `ArbitrageService`, asserted by both. Flagging it so it doesn't read as accidental inconsistency — and the general hazard is worth internalizing: *any* guard implemented at the REST boundary is invisible to Sidekick, because Sidekick is not an MCP client and does not speak HTTP.

**No tool declares an `owner` parameter, and none should.** This extends the existing `show_component` rule (the model never picks whose data) from component props to tool arguments. A model-supplied `owner` means a hallucination, since no schema declares one, and is dropped at dispatch.

### The docs half

The matrix documented **5 servers / 51 tools** against an actual **7 / 62**, omitted two servers entirely, and asserted a dual-registration that doesn't exist in code — `60e4bcd` moved every options tool off stock-price, which is the whole of the 29→21 drop. Counts are now taken with the **anchored** `grep -c "^@mcp.tool" fastMCPTest/*.py`; the unanchored form reports 63 because `options_analysis.py` mentions `@mcp.tool()` in its module docstring. That gotcha is written into the docs rather than left to be re-derived.

§5.5 said "never blanket-mirror the whole API", which reads as *discourage exposure* when the intent is *force a design discussion*. It now says most REST capabilities **should** reach a tool, that what's forbidden is the mechanical one-tool-per-endpoint mirror, and that every new REST capability owes one of two artifacts in the same PR: a designed tool, or a written decision that it stays REST-only. Silence stops counting as a decision.

## Verification

- Full backend suite: **1469 tests, `OK (skipped=5)`** (812s, against the test DB).
- Frontend chat suite: 34 passed. No frontend changes — no new component is registered.
- Handlers exercised in-process against real services: all three registered; caps reject **before** `ArbitrageService` is touched; `get_portfolio_summary` → `{positions, totals}`, `get_symbol_lots` → `{symbol, lots, lot_count}`.

**Not verified, and it's the interesting part.** The model-driven E2E needs a real Anthropic key through the browser vault. The in-process check proves the handlers and caps work; it does **not** prove the model now *reaches for* the new tools with `SYSTEM_PROMPT` naming them — which is the actual hypothesis behind #208. Worth a manual pass: "find cointegrated pairs among MSTR, COIN, MARA", "how's my portfolio doing?", and an unlinked principal (portfolio question should apologize while price questions in the same turn still work).

## Scoped out

The ⭐ "Built But Not Yet Surfaced" section was **not** re-audited. Rows provably closed by the `/fundamentals` page (#147 Part D) are corrected; the rest carries a dated caveat saying it was last verified 2026-07-20. A stale-but-labelled list beats both a confidently wrong one and an unscoped re-audit bolted onto this PR.

Plan and checkpoint log — including the three things that misled me during the docs pass — in `docs/proposals/sidekick-tool-audit-plan.md`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -446,7 +446,10 @@ analysis on test, treat prod as read-only" operating rule — now that changes s
 (`deploy.yml` → test, `prod-rollout.yml` → prod), prod is the live system everyone reads from. The
 deployed `quantui` UI and the `.mcp.json` AI-client remotes both already target prod.
 
-The 5 remote MCP servers in `.mcp.json` send `Authorization: Bearer ${QUANTCORE_MCP_TOKEN}`, which
+The 7 remote MCP servers in `.mcp.json` (stock-price, company-fundamentals, options-analysis,
+portfolio, arbitrage, news-sentiment, market-analysis — carrying **62 tools**; count them with the
+**anchored** `grep -c "^@mcp.tool" fastMCPTest/*.py`, since the unanchored form counts a mention
+inside `options_analysis.py`'s module docstring) send `Authorization: Bearer ${QUANTCORE_MCP_TOKEN}`, which
 the wrappers forward unchanged to `quantcore-api` (identity passthrough → the legacy HS256
 service-token path in the now dual-mode `api/auth.py`). So real analysis requires `QUANTCORE_MCP_TOKEN` to be a valid prod JWT in the
 environment Claude Code launches from; if it's unset, every data tool returns `401: … Not enough
--- a/api/auth.py
+++ b/api/auth.py
@@ -263,3 +263,39 @@ def require_owner(principal: Principal = Depends(require_principal)) -> str:
             datetime.now(timezone.utc).isoformat(),
         )
         raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="not_provisioned")
+
+
+def resolve_owner_or_none(principal: Principal) -> str | None:
+    """Resolve a principal to its canonical owner, or ``None`` if unmapped.
+
+    The soft sibling of :func:`require_owner`, for a request that is only
+    *partly* owner-scoped. A Sidekick chat turn serves many tools, most of which
+    read nobody's private data; 403-ing the whole turn because the caller's
+    portfolio identity is unprovisioned would kill ``get_stock_price`` along with
+    ``get_portfolio_summary``. So the resolution is soft here and the *tool call*
+    fails closed instead (``ChatService`` refuses to touch ``PortfolioService``
+    with a ``None`` owner) — the degrade is per capability, not per request.
+
+    Not a ``Depends``: it takes the principal a route has already resolved, so a
+    route can thread the answer into a context object rather than a parameter.
+
+    Every other property of ``require_owner`` is deliberately identical — the
+    local exemption, and above all **never auto-provisioning** (issue #126
+    decision #2). Softening the failure must not soften the mapping rule; the
+    only difference is that the caller gets to decide what the miss costs. The
+    security event is logged under a **distinct** name so the soft path stays
+    separable from the hard 403 in monitoring.
+    """
+    if principal.is_local:
+        return "john"
+
+    try:
+        return get_services().identity.resolve_owner(principal.owner)
+    except UnknownIdentityError:
+        # Identity only — never the token, header, or raw claims.
+        _security_logger.warning(
+            "SECURITY unmapped_principal_soft identity=%s at=%s",
+            principal.owner,
+            datetime.now(timezone.utc).isoformat(),
+        )
+        return None
--- a/api/routers/arbitrage.py
+++ b/api/routers/arbitrage.py
@@ -20,14 +20,19 @@ from typing import Optional
 
 from fastapi import APIRouter, HTTPException, Query
 
+from quantcore.services.arbitrage import (  # noqa: F401  (re-exported for callers/tests)
+    MAX_DISCOVER_REFERENCES,
+    MAX_DISCOVER_SYMBOLS,
+)
+
 from ..deps import services
 from ..json_response import QuantCoreJSONResponse
 
 router = APIRouter(prefix="/api/arbitrage", tags=["arbitrage"])
 
-# Each swept symbol costs a history fetch plus a ticker_info lookup.
-MAX_DISCOVER_SYMBOLS = 25
-MAX_DISCOVER_REFERENCES = 10
+# Each swept symbol costs a history fetch plus a ticker_info lookup. The
+# numbers live on the service module so this route and the Sidekick tool
+# handler share one definition (issue #208).
 
 # Below ~30 daily bars the service reports "no price history" anyway; the
 # ceiling is a sanity bound, well past the gateway's 730-day fetch cap.
--- a/api/routers/chat.py
+++ b/api/routers/chat.py
@@ -9,6 +9,11 @@ sealed envelope/scope plus the caller's principal, and hands it to the
 service opaquely. ``require_principal`` here is the same dependency the app
 already applies router-wide, so FastAPI's per-request dependency cache means
 the token is verified once, not twice.
+
+The context also carries a **soft-resolved** owner (issue #208), which is why
+this route does not depend on ``require_owner``: most chat tools are not
+owner-scoped, and one unprovisioned identity should cost the caller the
+portfolio tools, not the whole turn.
 """
 from __future__ import annotations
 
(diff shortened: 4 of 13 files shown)


JohnFunkCode (commented):
## Hermes Agent PR Review

**Reviewed head:** `6a7f4e8b961067acb7584321626a2b4b61677d75`
**Verdict:** COMMENT

### Code findings

No blocking code, security, performance, or test-quality findings identified in the reviewed diff. The owner-scoping path fails closed for unmapped identities, strips model-supplied `owner` arguments, and the discovery handler re-applies the REST caps before calling the service.

### Associated issue: #208 — Research: Sidekick doesn't have access to all MCP server tools — audit against capabilities-matrix.md and update it

- **Fully met:** The MCP inventory and counts are brought to the verified 7-server/62-tool state; the stale Sidekick prompt now names the relevant tool categories and delegates component details to the schema; the two concrete accidental gaps identified by the PR (`discover_arbitrage_pairs` and portfolio prose tools) are implemented; the capabilities matrix and architectural-standard documentation are updated; owner isolation and discovery-cap behavior have meaningful regression tests.
- **Partially met:** The issue asks for a determination for *each* MCP tool absent from Sidekick, including whether it should be added or remain excluded and the reason. The PR documents the broad policy and the three additions, but it does not provide an exhaustive per-tool disposition/recommendation for the remaining omitted MCP tools. The matrix's blank Sidekick cells show omissions but generally do not explain whether each is deliberate, unsuitable for chat, redundant, or simply deferred.
- **Unmet:** None conclusively identified.
- **Unverifiable:** The PR's model-driven E2E hypothesis — that the updated system prompt causes the model to reach for the new tools — was explicitly not verified. The in-process tests do not establish that behavior.

### Verification

- GitHub checks: **pass** — Tests + wrapper smoke + OpenAPI surface; Frontend tests + typecheck + coverage; Lean install import smoke; Secret scan (gitleaks).
- GitHub checks skipped: Cloud Run deployment and deploy-credential checks (workflow-reported as skipped).
- Local `python -m compileall -q api quantcore tests`: **pass**.
- Local `git diff main...HEAD --check`: **pass**.
- Targeted local unittest command was attempted but could not complete because this review environment lacks the `pandas` dependency (`ModuleNotFoundError` in the constant-parity test). The passing GitHub backend check is the applicable full-suite evidence.

The implementation itself looks sound; resolving the exhaustive missing-tool disposition and, if possible, performing the documented model-driven E2E checks would make the associated issue assessment complete.

PR #210, opened 2026-08-15
fix: five defects the post-#209 Sidekick test pass surfaced

JohnFunkCode:
Follow-up to #209 (issue #208). Five defects the post-merge Sidekick test pass on
**test** surfaced. Only the first was caused by #208; the middle three predate it,
and the last is an omission #208's new tool made visible. They are unrelated fixes
batched into one PR by request.

## 1 — A prompt rule that beat its own carve-out (`quantcore/services/chat.py`)

The reported symptom. The rule read:

> Most components are single-ticker: when the user asks about several symbols at
> once, give the numbers in prose instead, unless a component already covers the
> whole set (…).

The dominant clause is an instruction; the exception is a trailing subordinate
clause. Asked to sweep 32 tickers, the model gave prose and never rendered the
discovery card the sweep exists to feed — the exception lost. This is structurally
the same failure as known issue #121, so the fix inverts the sentence rather than
adding emphasis to the exception: lead with the set-taking components by name,
make prose the *fallback*, and say when the fallback applies.

## 2 — A risk card that mixed two pricing bases (`SpreadPayoffCard.tsx`)

`const debit = data.mid_debit || data.debit` put the **mid** debit in the headline
chip while `max_profit` / `max_loss` / `breakeven` / `R:R` beside it are all derived
server-side from the **conservative** debit (long ask − short bid). So the card
could read `Debit: $4.80 · Max loss: $4.94` — the one invariant a reader checks on
a vertical (max loss *is* the debit) visibly broken, with the smaller of the two
numbers sitting where "what you pay" goes.

The backend is self-consistent and unchanged: it deliberately draws the payoff
curves at mid (`curves.params.debit`) while keeping the scalars conservative. The
card now shows the conservative debit in the headline chip, the mid as its own
labelled chip when it differs by ≥ half a cent, and notes the curve basis in the
caption. `mid_debit` was also typed `number` when the server returns `null` if
either leg lacks a mid — which is what made the `||` look harmless; it is now
`number | null`.

## 3 — The economic-link gate false-positived on "mining" and "energy" (`arbitrage.py`)

The gate is what stops a wide statistical sweep from manufacturing cointegrated
pairs. It was a bare substring test against `sector + industry + summary[:400]`,
with generic industry words in `REFERENCE_PANEL` (`mining`, `energy`, `materials`,
`technology`, `software`). Two distinct false links came out of that:

- a **bitcoin** miner reads as "mining" and was cointegration-tested against gold
  and copper; "software"/"technology" on BTC-USD made every tech company eligible
  to be paired with bitcoin;
- **"Goldman"** contains "gold", so an investment bank was tested against gold
  futures.

Every keyword now names *the commodity itself*, matched word-boundary anchored,
with a trailing `*` marking a stem for the cases where the boundary is the thing
in the way (`crypto*` → "cryptocurrency", `digital asset*` → "assets").

Measured against live yfinance profiles over a 28-name sample before and after:
gate-passing (symbol, reference) pairs fell **72 → 29**, with no true positive
lost. MARA `GC=F,HG=F,CL=F,NG=F,BTC-USD` → `BTC-USD`; RIOT `GC=F,HG=F,BTC-USD` →
`BTC-USD`; AAPL/MSFT/NVDA `BTC-USD` → none; GS `GC=F,BTC-USD` → none; SLB
`CL=F,NG=F,BTC-USD` → `CL=F`. Gold/silver miners kept `GC=F`, FCX kept `HG=F`,
XOM/OXY kept `CL=F,NG=F`, MSTR and COIN kept `BTC-USD`.

Two candidate keywords were dropped *because* of that measurement, not review:
`refin` hits EXK and CDE (*refined* silver, which would link silver miners to
crude) and real refiners already match on `oil`; `drilling` is standard mining
exploration language.

## 4 — Holding periods were measured in UTC, not market time (`portfolio.py`)

`as_of = date.today()` is the *host's* date. On Cloud Run that is UTC, so from
roughly 5pm ET until midnight ET it is already tomorrow: a lot bought eight days
ago reports nine, and any per-day rate divided by it is understated by the same
ratio (~11% on a nine-day hold). Wrong numbers for seven hours of every day, and
never visibly wrong.

New `quantcore/analytics/market_time.market_date()` returns the Eastern-time date.
Deliberately **not** `latest_completed_session()` — that rolls back over weekends
and overnight, which is right for "is there a newer bar?" and wrong for a holding
period that keeps accruing. Applied to the three verified sites: `list_lots`,
`symbol_rows`, `portfolio_totals`.

**Scoped out on purpose:** 12 other `date.today()` sites exist (options.py
397/752/1082/1179, fundamentals.py 337/772, options_screening.py 543, arbitrage.py
137/454/700, options_position_repository.py 140/203). They are DTE and cache-age
calculations carrying the same drift, but I did not verify each one's semantics, so
rewriting them silently would be guessing. Worth a follow-up issue.

## 5 — `discover_pairs` stated completeness it could not support (`arbitrage.py`)

The return named only the symbols that *found* something. An agent splitting 32
tickers across two calls therefore had no way to reconcile its answers against the
input — in the test run it dropped SSRM and then reported full coverage. The
omission was undetectable from the response, so a confident wrong answer was the
only answer available to it.

`symbols_tested` and `skipped` now **partition** `symbols_requested`, each skip
carrying its reason. The reason matters as much as the accounting: a gate-blocked
symbol was never run through a cointegration test, so reporting it as "no
cointegration found" would be a second misstatement. Symbols are de-duplicated and
normalized on the way in. Shaped for the model rather than mirrored off the
endpoint (the §5.5 principle from #208): the tool description and the MCP wrapper
docstring both tell the agent to read those lists and to reconcile across calls.

`GET /api/arbitrage/discover` has no `response_model`, so the new keys pass through
additively; the frontend type carries them as optional.

## Tests

- `tests/test_arbitrage_service.py` — `EconomicLinkGateTest` (bitcoin miner links
  only to bitcoin; gold miner still links to gold; "Goldman" is not gold; stems
  reach their inflections) and `DiscoveryCoverageTest` (the partition holds; skips
  carry the right reason; a gate-blocked symbol is not counted as tested;
  duplicates collapse; `count` is out of tested, not requested).
- `tests/test_market_time.py` — `TestMarketDate`, including the 23:35 ET case where
  UTC is already tomorrow, and two contrasts against `latest_completed_session`.
- `SpreadPayoffCard.test.tsx` — the existing assertion literally encoded the bug
  (`toContain('4.80'); // mid debit`) and was corrected; five new cases cover the
  chip split, `mid_debit: null`, mid == conservative, and `mid_debit: 0` (the case
  the `||` swallowed).

Backend suite green; frontend 613 tests / 82 files green; `tsc --noEmit` clean.


JohnFunkCode:
## Hermes Agent PR Review

**Reviewed head:** `643c759d337ee827fe678a3aeb67e672157ffac7`
**Verdict:** APPROVE

### Code findings
No Critical, Warning, or Suggestion findings. The five fixes are scoped coherently, the conservative-vs-mid debit distinction is reflected in both implementation and tests, the economic-link regex changes address the documented false positives, and the coverage partition is tested for skips, deduplication, and gate behavior.

Issue #208 — *Research: Sidekick doesn't have access to all MCP server tools — audit against capabilities-matrix.md and update it* — is referenced by this PR as the already-completed parent/follow-up context. Its objectives were fulfilled by PR #209, which closed the issue; this PR does not claim to modify that audit and no remaining #208 objective is left for this follow-up.

- **Fully met:** #208 objectives are already fulfilled in the repository by merged PR #209; this PR's five follow-up defects are implemented in the diff.
- **Partially met:** None.
- **Unmet:** None.
- **Unverifiable:** None.

### Verification
- GitHub checks for this head: **pass** — frontend tests/typecheck/coverage, backend tests/wrapper smoke/OpenAPI surface, lean import smoke, and secret scan.
- Local targeted execution could not run because this review environment lacks the repository Python dependencies (`pytest`, `numpy`, `pandas`) and frontend `node_modules`; the authoritative GitHub checks passed.
- No existing review or review comment for this head was found before posting.

*Formal approval was unavailable because the authenticated user is the PR author; this top-level review comment is the fallback required by the repository review procedure.*

